PR #24, opened 2014-11-06
Editors picks index

janua:
We want a front to say what `collection` on itself is considered `editorsPicks`.

We are going to use an `index` instead of a `collectionId` to avoid for when a collection falls out of the path, it would continue to be used.

In the case of a non-existent index, or no value, we will default to the top collection, which is the current behaviour. (This is all going to be in `frontend`)

@stephanfowler @robertberry


janua:
This is now irrelevant, and should be replace with the new `canonical` implementation.

PR #31, opened 2014-12-29
Fapi client integration

adamnfish:
Here's a rough working FAPI-client.

Note that these changes are based on those in https://github.com/guardian/facia-scala-client/pull/26, viewing the changes since that branch (instead of master) will reveal the changes that I have made beyond what @janua has done.

The `FAPI` class contains the main interface to the client library. It includes methods that fetch info from Facia and the Content API. Facia's logic is applied to this information before being returned in data types convenient for consumption by other applications.

I'm raising this now so you all see it working, there's still more to do.
e.g. For now we only support standard content trails in the curated content. Some work will need to be done to resolve the various types of snaps into curated content instances.


README.markdown
```diff
@@ -1,20 +1,45 @@
...
 ==================
 
-Low-level client for Facia JSON API.
+Facia's Scala client is split into two parts.
 
-Usage
------
+## The Fronts API client
+
+The `fapi-client` project contains the main Fronts API client. This provides ways to fetch Fronts
```
robertberry: :crying_cat_face:

build.sbt
```diff
@@ -65,3 +52,38 @@ ReleaseKeys.releaseProcess := Seq[ReleaseStep](
...
+    scalaVersion := "2.11.4",
+    crossScalaVersions := Seq("2.10.4", "2.11.4"),
+    organization := "com.gu",
+    name := "fapi-client",
+    resolvers ++= Seq(
+      Resolver.file("Local", file( Path.userHome.absolutePath + "/.ivy2/local"))(Resolver.ivyStylePatterns),
+      "Typesafe Repository" at "http://repo.typesafe.com/typesafe/releases/"
+    ),
+    libraryDependencies ++= Seq(
+      "com.gu" %% "content-api-client" % "4.0-SNAPSHOT",
```
robertberry: Should probably be consistent with how dependencies are defined above (vals from a separate Dependencies config).

@rtyley showed me this originally, but it's a really good way of easily seeing all the dependencies of a project, but also making sure versions don't become out of sync between subprojects

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.gu.facia.api.contentapi.ContentApi
+import com.gu.facia.api.contentapi.ContentApi.{AdjustItemQuery, AdjustSearchQuery}
+import com.gu.facia.api.models._
+import com.gu.facia.client.ApiClient
+import com.gu.facia.client.models.{CollectionConfigJson, TrailMetaData}
+
+import scala.concurrent.ExecutionContext
+
+
+object FAPI {
```
robertberry: As `capiClient` and `faciaClient` are required for every method here, why is this not just a `class` where those are passed in as parameters on construction?

fapi-client/src/main/scala/com/gu/facia/api/Response.scala
```diff
@@ -0,0 +1,60 @@
+package com.gu.facia.api
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+case class Response[A] private (underlying: Future[Either[ApiError, A]]) {
```
robertberry:
```scala
import scalaz._
import Scalaz._

type EitherApiError[A] = Either[ApiError, A]
type Response[A, B] = FutureT[EitherApiError, A, B]
```

:trollface:

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -0,0 +1,89 @@
...
+import com.gu.facia.api.{CapiError, Response}
+
+import scala.concurrent.ExecutionContext
+
+
+object ContentApi {
+  type AdjustSearchQuery = SearchQuery => SearchQuery
+  type AdjustItemQuery = ItemQuery => ItemQuery
+
+  def buildHydrateQuery(client: GuardianContentClient, ids: List[String]): SearchQuery = {
```
robertberry: Do we do any tests for maximum URL size here?

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -0,0 +1,83 @@
...
+    collection.live.flatMap { trail =>
+      val contentCodeUrl = trail.id
+      content.find(c => trail.id.endsWith("/" + c.safeFields.getOrElse("internalContentCode", throw new RuntimeException("No internal content code")))).map { content =>
+        FaciaContent.fromTrailAndContent(content, trail.safeMeta, collectionConfig)
+      }
+    }
+  }
+}
+
+sealed trait Group
```
robertberry: Is this useful to have?

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
```diff
@@ -0,0 +1,119 @@
...
+
+import com.gu.facia.client.models.CollectionConfigJson
+
+case class Groups(groups: List[String])
+
+sealed trait DynamicContainerType
+object DynamicFast extends DynamicContainerType
+object DynamicSlow extends DynamicContainerType
+
+sealed trait FixedContainerType
```
robertberry: Not sure about these either. I think a string ID (maybe wrapped in a type) would be fine.

robertberry:
Looks good otherwise I think :+1:

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
adamnfish: a) because they aren't required everywhere
b) because this way we just have pure functions. Whenever you have code that has a dependency on some other value, there are lots of ways of getting that value in scope. People always end up making traits and classes and so on to get that to happen. But you know how else you can make code in a function depend on a value? With an argument. It's amazingly simple and I'm always baffled people don't just do this everywhere.

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
adamnfish: That's an excellent point. No, we're likely to run into problems with that at some point.

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
adamnfish: The Group bit, you mean? I think so...

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
adamnfish: @janua did mention he wasn't sure about these bits, probably worth chatting with him?

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
robertberry: Why not just a type wrapping the integer value, though? That might change in the future, and the way we both deal with them is at an integer level.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
robertberry: I don't mind using arguments as dependencies, but I think `implicit`s are a bit black magic, and should only really be used for typeclasses or the 'pimp my library' pattern (I wish there was a different name for that, I hate saying it!).

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
adamnfish: If memory serves @janua reckoned this was what it *should* be - this lib is supposed to turn the dirty internals into useful stuff that properly describes the data types.

robertberry: Ja, but I think the data type is just the integer value. If you make it into a sum type here you lose all ordering. (You don't know that Standard < Big unless you explicitly code that in yourself, which is silly - the integer gives you that to begin with.)

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
robertberry: I'll just change it to a wrapper for the string ID. You're not using these at the moment, and I'd prefer to just have the string available in `frontend`.

robertberry:
@adamnfish I've added id batching. pls check over

adamnfish:
@robertberry 
No need to bump the version if you're just going to add to this branch. Really, we should try and add new stuff in another branch, which might mean merging this? I have another change to add now as well.

It'd be good to have an integration test that does the batching as well, so we can be confident it works against the Content API properly and so it can't be inadvertently broken later.

robertberry:
I don't think I did bump the version :/ Yes, let's merge this.

I can't get the integration tests to run. It keeps complaining that I don't have the right credentials for some reason ...

PR #32, opened 2015-01-02
Don't flake out on 404s

robertberry:
If a collection has never been edited, its collection JSON won't exist. Clients ought to behave properly in that situation, as everything can be inferred from the config.


adamnfish:
:+1:

PR #33, opened 2015-01-02
Add collections for front endpoint

adamnfish:
This adds a new endpoint that lets you grab all the collections for a front in one go. This is essential for being able to render a front, so it's probably (over the existing single collection endpoint) the main entry point for getting collection info.


janua:
:+1:
Looks good; will this cause a big hit to the CAPI if there were a lot of collections?

adamnfish:
Nah, this doesn't hit CAPI, just loads the Facia stuff. You need to call collectionContent to do the CAPI stuff.

PR #34, opened 2015-01-05
Extract some Front logic that was embedded in the FAPI file

adamnfish:
This allows Mobile to use the logic without duplicating it and improves testability.


janua:
:+1:

PR #35, opened 2015-01-05
Add TestContentApiClient

janua:
This is to allow overriding of targetUrl from environment variables.

It's only for `IntegrationTestConfig`.

@robertberry @adamnfish


robertberry:
:+1:

PR #36, opened 2015-01-05
Upgrade to 4.1 of content-api-scala-client

janua:
@robertberry @adamnfish


janua:
Should we add Sonatype to this project instead of relying on local configurations?

robertberry:
I dunno, I think not as it is just supposed to be a staging repository. :+1:

PR #37, opened 2015-01-05
Reduce parallelism to prevent AWS SDK failures

adamnfish:
This fixes failures in our integration tests. With these calls made in parallel we consistently (ish) get 404s from the S3 calls. By swapping the concurrent (traversed) calls over to serial ones (sequence), these problems go away. The downside is that calls to retrieve collections will take much longer.

@robertberry, @janua is there a better way you can think of with some batching / configuration of the AWS SDK?

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -52,7 +52,7 @@ object FAPI {
       configJson <- Response.Async.Right(faciaClient.config)
       frontJson <- Response.fromOption(configJson.fronts.get(frontId), NotFound(s"No front found for $frontId"))
       collectionIds = frontJson.collections
-      collectionsJsons <- Response.Async.Right(Future.traverse(collectionIds)(faciaClient.collection))
+      collectionsJsons <- Response.Async.Right(Future.sequence(collectionIds.map(faciaClient.collection)))
       collectionConfigJsons <- Response.traverse(
         collectionIds.map(id => Response.fromOption(configJson.collections.get(id), NotFound(s"Collection config not found for $id")))
       )


robertberry:
404s? That seems like a weird error to be getting?

adamnfish:
IKR

adamnfish:
com.amazonaws.AmazonClientException: Unable to unmarshall error response (org.apache.xerces.parsers.XIncludeAwareParserConfiguration cannot be cast to org.apache.xerces.xni.parser.XMLParserConfiguration). Response Code: 404, Response Text: Not Found

janua:
@adamnfish What's going on with this? This seems like a weird error.
We have batching code in `frontend` used in this situation, and from what I remember it's quite easy to use.

adamnfish:
@janua sounds like that batching code would be useful in the Facia client then?

janua:
@adamnfish Yes, it would. I will port it over eventually.

What shall we do with this? Change it to sequence or leave it as traverse?

robertberry:
I don't see why changing to `sequence` would help here, unless something that isn't threadsafe is being used on each element in the list.

If that's the case then we should probably fix that instead as it could lead to other weird bugs.

adamnfish:
* changing to sequence literally does fix the problem
* the problem comes from the AWS SDK's internals, which we know is terrible

The integration tests prove that it's easy to make this fail with traverse in place, so I think we have to switch it to sequence? Obviously that's *far* from ideal, so this is something that needs a bit of study.

janua:
@adamnfish What ever happened here? This could still be an issue for me for when I merge into `frontend` fully. I will leave this open and close it in the case of not seeing any issues.

Has `MAPI` had any concurrency issues since implementation?

johnduffell:
I guess the issues didn't happen!

PR #38, opened 2015-01-06
Various Changes

janua:
This has various changes:
 - Removal of `CollectionId`
 - Addition of `lastUpdated`
 - Changing of `updatedBy` and `updatedEmail` to `String` instead of `Option[String]`

I have also changed `Collection.fromCollectionJsonConfigAndContent` to take a `CollectionJson` instead of `Option[CollectionJson]`. The adding of `lastUpdated` really highlighted the fact that this probably should be taking a solid `CollectionJson`; and it also seems wrong to be constructing `Collections` without the actual `CollectionJson`. This isn't ideal though, as you can have `CollectionConfig` without the `CollectionJson` yet have been created. The extra thing we could do is to have it fallback to an empty `CollectionJson`, but what does an empty `CollectionJson` look like?

@adamnfish @robertberry 

I forgot to mention, an extra commit sneaked in by accident; the removal of snaps from the request for the content api.


fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -52,23 +53,24 @@ object FAPI {
       configJson <- Response.Async.Right(faciaClient.config)
       frontJson <- Response.fromOption(configJson.fronts.get(frontId), NotFound(s"No front found for $frontId"))
       collectionIds = frontJson.collections
-      collectionsJsons <- Response.Async.Right(Future.traverse(collectionIds)(faciaClient.collection))
+      collectionsJsonOptions <- Response.Async.Right(Future.traverse(collectionIds)(faciaClient.collection))
+      collectionsJsons = collectionsJsonOptions.flatten
```
adamnfish: This won't work.

If you look below, you'll see these lists get zipped together. This only worked on the assumption they were the same length.

:-1:

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -52,23 +53,24 @@ object FAPI {
...
-        Collection.fromCollectionJsonConfigAndContent(CollectionId(collectionId), collectionJson, collectionConfig)
+        Collection.fromCollectionJsonConfigAndContent(collectionId, collectionJson, collectionConfig)
       }
     }
   }
 
   def collectionContent(collection: Collection, adjustSearchQuery: AdjustSearchQuery = identity)
                        (implicit capiClient: GuardianContentClient, ec: ExecutionContext): Response[List[CuratedContent]] = {
-    val itemIds = collection.live.map(_.id)
+    val itemIdsForRequest = collection.live.filterNot(_.isSnap).map(_.id)
```
adamnfish: This logic should probably be elsewhere if possible. Better to keep logic out of the FAPI file because it can't be tested here and I think this is something that needs a test (if only to prevent regressions).

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -61,7 +61,7 @@ object FaciaContent {
 
     CuratedContent(
       content,
-      trailMetaData.headline.orElse(contentFields.get("headline")),
```
adamnfish: Are we sure we don't need to favour `contentFields.headline` over `content.webTitle`?

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
```diff
@@ -73,12 +74,13 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
 
   "backfill" - {
     val collection = Collection(
-      CollectionId("uk/business/regular-stories"),
+      "uk/business/regular-stories",
       "economy",
       Nil,
       None,
-      Some("updatedBy"),
-      Some("updatedBy@example.com"),
```
adamnfish: Talk to @robertberry about this. He literally only just switched them over to Options and now you're switching them back.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
adamnfish: Remember that not all collections have collection JSON, so it really does need to be an option.

adamnfish:
Can you split this PR up?

* The date change is fine
* The filtering snaps change looks ok ish, but needs testing (so the logic should probably move)
* the 'removing Option' change doesn't work as is

robertberry:
@janua they need to be Options, as if the container hasn't yet had an item put into it (such as containers that are only backfills ... ), those values won't be set

janua:
@robertberry This is what we have been talking about, have a default CollectionJson or something else to deal with this. They *should* always be set. In fact if they are not set, it won't parse.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
janua: To bring the conversation we had in here; this is what the defaults would be for that I mentioned. So we either change it back to an `Option` and then have the resolving of the default fields in the creation of `Collection`, or we keep this and change the zipping and use A default `CollectionJson`.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
janua: I'll move this to `Collection`.

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
janua: I have asked @maxharlow about this; it seems the fields work like this:
`fields.headline` is for the headline within the article, where `webTitle` is the headline for the link name on the front. We happen to have out `CuratedContent` called `headline`, so I think this is probably correct. Maybe we should change `CuratedContent.headline` to `webTitle`?

adamnfish:
@janua totally up to you guys, but either way please can you split this PR up? I need the date change while you work out what's best for the collection stuff. This whole "not parse" thing isn't true though. It parses just fine as it is, it's only your changes that have created that problem?

janua:
@adamnfish I will not split up this, rather I will create a new PR now and cherry pick what I need from here. And resolve this if anything conflicts when merging.

robertberry:
@janua Hmm right I see what you mean. I guess the situation in which they're not set, the collection JSON doesn't exist, you only have the backfill ...

janua:
@robertberry Which would mean... they everything that comes from `CollectionJson` either has to be `Option` or a default somehow.... (I was trying to avoid the `Options`)

robertberry:
Some probably have sensible defaults, but last updated by doesn't if it's never been editorially curated

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
maxharlow: I've checked this, and I think it's slightly different.

The Oracle database fields used:

`webTitle` --> `page_live.link_text`
`fields.headline` --> `content_live.headline`

(as seen here: https://github.com/guardian/api-indexer/blob/master/src/main/resources/sql/content-query-template.sql)

Most of the time they are the same, [but sometimes not] (http://content.guardianapis.com/lifeandstyle/wordofmouth/2015/jan/06/-sp-cup-soup-instant-taste-test-batchelors?api-key=test&show-fields=headline) -- the webTitle is used for the `<title>` (presumably there's some SEO benefit), and the headline for, uh, the headline. Not sure about fronts -- it's currently on the network front with a completely different (much shorter) title. On the `lifeandstyle/wordofmouth` tag page it appears to use either the headline, of another field set to exactly the same value.

janua: @maxharlow Thanks!

@adamnfish Maybe to make this concrete we should use `MetaData.headline` -> `fields.headline` -> `content.webTitle`?

janua:
@adamnfish @robertberry 
All this PR really is now is:
 - Removal of `CollectionId` miss you :cry: 
 - Addition of `liveIdsWithoutSnaps`

With the pending decision on the headline.

robertberry:
Why are you getting rid of `CollectionId`, btw? It seemed fine to me.

Looks good though

adamnfish:
CollectionId is a pain. It doesn't really offer any extra safety, is inconsistent with every other type in the library, and makes the code in applications that use this lib all the more unwieldy (since they only care about the strings). If we consistently apply AnyVals and name things clearly then it could be OK, but this wasn't so I asked for this, sorry.

janua:
@robertberry 
@adamnfish didn't like it, and he's the first user of the client

/ grabs popcorn

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
adamnfish: Yep, that's exactly what I'm doing in MAPI now, be good to get that into the lib so we don't have to do the same thing over and over.

robertberry:
Meh, you have to start somewhere. I value things gradually evolving to be better more than a completely consistent codebase. But anyhow ...

janua:
@adamnfish @robertberry If you are happy could you plus one it :smile:

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
janua: @adamnfish Done

robertberry:
:+1:

PR #40, opened 2015-01-06
Added a proper recover fn - old recover was mis-named and has been fixed

adamnfish:
Recover now exists to provide an alternative to a failure. The old recover has been named mapError, since that's actually what it does (it's actually analogous to Either's `.left.map`).


janua:
:+1:

PR #44, opened 2015-01-06
Fix Sonatype stuff

robertberry:
Makes it so you can release either client ... 

Although I'm not sure of the best way to manage the version upping and tagging (as it does it for each subproject, when really you want to do a release for each, but only up the version once). I've copied @steppenwells for this, maybe he has an idea ...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,53 +8,65 @@ name := "facia-api-client"
 
 scalaVersion := "2.10.4"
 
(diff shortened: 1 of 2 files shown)


robertberry:
CC @rtyley

janua:
@robertberry 
Is it possible to get this merged and a version released?
Maybe we should include #45 before we release.

PR #47, opened 2015-01-07
Fix flakey test

janua:
@robertberry


robertberry:
Do all articles not have types either? Sigh. We need an HTTP recorder.

:+1:

janua:
Seems not, although I imagine it is rare.

:telephone_receiver: Hello?

PR #48, opened 2015-01-08
Snaps

janua:
This introduces Snaps to the Fapi client.

In the ADT, `Snap` hangs off `FaciaContent`, while `LinkSnap` and `LatestSnap` hand off `Snap.

The small changeset:
 - Ability to request content with or without snaps through `collectionContentWithSnaps` and `collectionContentWithoutSnaps`
 - Can adjust the default `LatestSnap` query

The behaviour is that if a `LatestSnap` gets no results from the `ContentAPI`, it will just stay as a `LatestSnap` with its `latestContent` set to `None`.
The same behaviour applies to requesting content for a `Collection` via `...withoutSnaps`; they stay as `LatestSnap` with an empty `latestContent`.

@robertberry


robertberry:
:+1:

PR #50, opened 2015-01-09
Fix datetime

janua:
@robertberry Why in `package`? :cry:


robertberry:
Can we just delete @adamnfish 's silly tests?

:+1:

PR #54, opened 2015-01-14
Add previously to CollectionJson

janua:
`previously` is used to show what has been removed and previously existed in the collection.

It is generally capped to `20`.

@robertberry @adamnfish


robertberry:
:+1:

PR #57, opened 2015-01-26
Supporting Items

janua:
This adds the requesting of `supporting` items.

When requesting `snaps`, it fills out `dreamsnaps` inside `supporting` items.

@robertberry


robertberry:
:+1:

PR #59, opened 2015-01-28
Add filter to response

andrewcb:
This adds a filter method which, if unsuccessful, returns a response whose Left is a new ApiError subclass named FilteredOut.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/Response.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/Response.scala
@@ -15,7 +15,7 @@ case class Response[A] protected (underlying: Future[Either[ApiError, A]]) {
     }
   }
 
-  def filter(f: A => Boolean, orLeft: ApiError = FilteredOut())(implicit ex: ExecutionContext): Response[A] = flatMap { a =>
+  def filter(f: A => Boolean, orLeft: ApiError = FilteredOut)(implicit ex: ExecutionContext): Response[A] = flatMap { a =>
     if(f(a))
       Response.Right(a)
     else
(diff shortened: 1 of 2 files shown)


fapi-client/src/main/scala/com/gu/facia/api/Response.scala
```diff
@@ -81,3 +88,4 @@ case class DataError(message: String, cause: Option[Throwable] = None) extends A
 case class CapiError(message: String, cause: Option[Throwable] = None) extends ApiError
 case class HttpError(message: String, cause: Option[Throwable] = None) extends ApiError
 case class UrlConstructError(message: String, cause: Option[Throwable] = None) extends ApiError
+case class FilteredOut(message: String = "Filtered out", cause: Option[Throwable] = None) extends ApiError
```
janua: Can we change this to an `object`?

janua:
:+1:

fapi-client/src/main/scala/com/gu/facia/api/Response.scala
```diff
@@ -15,6 +15,13 @@ case class Response[A] protected (underlying: Future[Either[ApiError, A]]) {
     }
   }
 
+  def filter(f: A => Boolean, orLeft: ApiError = FilteredOut)(implicit ex: ExecutionContext): Response[A] = flatMap { a =>
```
adamnfish: This should probably be called withFilter so it can be used in for comprehensions? Although to do that canonically it would have to be implemented lazily, IIRC. Worth testing because the for-comp syntax would be the correct way of actually using this behaviour in most use-cases.

janua: Yeah agreed. I suggested pointing `withFilter` to `filter` for the time being until you figure out how to make it truly lazy.

PR #61, opened 2015-01-30
Add most-viewed results to the backfill query

adamnfish:
This fixes backfill queries for the "most popular" collections.

cc @janua


andrewcb:
:+1:

PR #63, opened 2015-01-30
Add covariance on Response

janua:
`Response.Right(Nil)` is not compatible with the type `Response[List[T]]`. This fixes that.

@adamnfish


adamnfish:
:+1:

PR #64, opened 2015-02-06
Make debugging backfill results simpler

adamnfish:
It's now possible to drop a debug statement on the results in IntelliJ, which makes debugging backfill issues *MUCH MUCH* easier. I've been using this in a publish-local all the time, so it's worth having in the real world as well.


janua:
:+1:

PR #65, opened 2015-02-06
Add Importance

janua:
This adds an `importance` field to the `CollectionConfigJson` type, and carries through the field into the type `Importance`.

@adamnfish @piuccio @stephanfowler

### DIFF
--- a/fapi-client/src/test/scala/com/gu/facia/api/models/ImportanceTest.scala
+++ b/fapi-client/src/test/scala/com/gu/facia/api/models/ImportanceTest.scala
@@ -5,13 +5,10 @@ import org.scalatest.{Matchers, FlatSpec}
 
 class ImportanceTest extends FlatSpec with Matchers {
 
-  def collectionConfigJsonWithImportance(importance: Option[String]) =
-    CollectionConfigJson.withDefaults(importance = importance)
-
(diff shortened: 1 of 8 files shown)


piuccio:
given that I don't understand scala :+1:

fapi-client/src/test/scala/com/gu/facia/api/models/ImportanceTest.scala
```diff
@@ -0,0 +1,32 @@
+package com.gu.facia.api.models
+
+import com.gu.facia.client.models.CollectionConfigJson
+import org.scalatest.{Matchers, FlatSpec}
+
+class ImportanceTest extends FlatSpec with Matchers {
+
+  def collectionConfigJsonWithImportance(importance: Option[String]) =
```
adamnfish: Is this indirection really adding anything? It'd be simpler to just use withDefaults directly, I think?

janua: I thought I was going to have to fill out the full signature when I first made it, that's why.

adamnfish:
Looks ok to me!

stephanfowler:
:+1:

PR #66, opened 2015-02-09
Carry MetaData for LinkSnap and LatestSnap

janua:
We are missing a lot of information for `LinkSnap`s and `LatestSnap`s that should be coming from `Metadata`.

This brings in all the `TrailMetaData` fields into both `Snap` types.

@adamnfish


robertberry:
:+1:

PR #67, opened 2015-02-12
Remove Duplication between CollectionConfig and Collection

janua:
#### What
This removes some of the repetition between `CollectionConfig` and `Collection`.

Instead, `Collection` now carries a `CollectionConfig`.

I have left the duplication for `href` and `displayName` until we completely remove these fields out of the `CollectionJson` and `Collection`.

#### Why
Duplication is too confusing and new fields don't get naturally carried unless explicit.

@robertberry @adamnfish


robertberry:
:+1: :cat:

PR #69, opened 2015-02-16
Image Cutout Logic

janua:
This changes `ImageCutout` on `CuratedContent` and `SupportingCuratedContent` to non-optional, and instead changes the fields to optional.

`ImageCutout.imageCutoutReplace` can be set to `true`, without there being an `ImageCutout` provided; this would mean that there should be an `ImageCutout` coming entirely from the tags of the content.

This copies some of the behaviour from `frontend`, but it is not as black and white there and I haven't completely taken everything across.

@adamnfish @robertberry

### DIFF
--- a/fapi-client/src/test/scala/com/gu/facia/api/models/ImageCutoutTest.scala
+++ b/fapi-client/src/test/scala/com/gu/facia/api/models/ImageCutoutTest.scala
@@ -34,65 +34,50 @@ class ImageCutoutTest extends FreeSpec with Matchers {
 
   "ImageCutout" - {
 
-    "should return empty ImageCutout" in {
-      ImageCutout.empty should be (ImageCutout(false, None, None, None))
-    }
-
-    "should return true for standalone true" in {
+    "should return None for imageCutoutReplace=true but nothing to replace with" in {
       val trailMeta = trailMetaDataWithImageCutout(true)
       val imageCutout = ImageCutout.fromContentAndTrailMeta(emptyContent, trailMeta)
-      imageCutout.imageCutoutReplace should be (true)
+      imageCutout should be (None)
     }
 
     "should return src with replace true or false" in {
-      val src = Option("src")
-      val width = Option("width")
+      val src    = "src"
+      val width  = Option("width")
       val height = Option("height")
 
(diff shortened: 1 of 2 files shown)


fapi-client/src/test/scala/com/gu/facia/api/models/ImageCutoutTest.scala
```diff
@@ -0,0 +1,99 @@
...
+      "webTitle", "webUrl", "apiUrl", None,
+      Nil, None, Nil, None)
+
+  def contentWithContributorTag(bylineLargeImageUrl: String): Content =
+    emptyContent.copy(tags = List(Tag(
+      "tagid", "contributor", None, None, "webTitle",
+      "webUrl", "apiUrl", Nil, None, None, bylineImageUrl = None,
+      Option(bylineLargeImageUrl), None, None, None, None)))
+
+  val bylineImageUrl = "http://static.guim.co.uk/sys-images/Guardian/Pix/pictures/2014/3/13/1394733744420/MichaelWhite.png"
```
robertberry: O__O

robertberry: michaelwhite

robertberry:
:+1:

janua:
:laughing: Nice transparency!

adamnfish:
I'm not sure this is entirely sensible. We're trying to model the actual data constraints in the type. Would it not be better to move the override boolean out of this and leave this as an overall option? The issue is, width and height presumably have the same "optionalness" as the image src, so modelling that properly is useful (and that's what it did before). Having nested models where everything is an option isn't useful! This library should be doing the resolution and making the decisions so that the data is clear?

janua:
Yes, the logic is weird.
We have two different scenarios:
 - `ImageCutout` from a `Tag` which has only got a `src` attribute and no `width` or `height`.
 - `ImageCutout` from `facia-tool` which gives a `src`, `width` and `height`.

Maybe we should have no boolean, and have only a solid `src` with optional `width` and `height`, which would be in turn inside an Option? The boolean would essentially be `isDefined` of the `boolean`.

janua:
@adamnfish Does that make more sense now?

PR #70, opened 2015-02-19
Carry all snapTypes through LinkSnap

janua:
We have different `snapType`s: `link`, `latest`, `document`, `fragment` and `json.html`.

`LinkSnap` is good enough to just carry these, but the name seems wrong so it will get renamed in the future, but not in the PR as it will clash with my massive PR on `frontend`. 

When changing the name, I may add proper types for each `snapType`. I'm not sure if it's hugely beneficial as we don't do different things for different types in scala, instead we just carry and output the `snapType`. (Apart from `LatestSnap`).

@adamnfish @robertberry


robertberry:
:+1:

PR #71, opened 2015-02-19
Added the kickerDisplayText function

andrewcb:


fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
```diff
@@ -89,6 +89,17 @@ object ItemKicker {
     case FreeHtmlKickerWithLink(body, _) => Some(body)
     case _ => None
   }
+
+  /** Return the canonical display text for a kicker */
+  def kickerDisplayText(itemKicker: ItemKicker): Option[String] = itemKicker match {
```
janua: What about `LiveKicker`? And can we call the method something else? It's too close to the method above.

adamnfish: Why have this function at all? Wouldn't it make sense to just use that one function for all of the types, instead of doing half here and half over there? It makes it hard for someone using the client to know which of these two functions is the right one to use.

janua: @adamnfish Because the other function returns `HTML`, which I believe you do not want. We also do our own mappings on `frontend`, so we have no need for this new function.

adamnfish: Yeah, I thought that might be the reason BUT. If we don't support HTML that's our problem so maybe it doesn't need to be in the Facia client (we could just strip it, unless the HTML is complex and meaningful - you'll know best on this score)?

If that's the case though (and you think this is helpful in the client) then the function above should probably be renamed to `kickerHTML` and this one should be `kickerText`.

janua: Yes, that naming is much better.

During my traverse for `displayName`, I am also going to have a look at what sort of things exist inside the kickers when they are `HTML`. Maybe we will be able to keep `HTML` out of it, but i'm not sure as I am not very familiar with `Kickers`, this is very much a `frontend` concept. Hopefully we can avoid any nasty `HTML` stripping.

I can't see what the problem is with resolving this text yourself though, I mean, you have the `types`! :smile: The only benefit is if we did start using it you would have consistency across app and web, but are we looking to have this? Design in the apps can be different to the design on the web!

adamnfish: It'd be good to understand exactly what is in those fields before we try and set the interface in stone. Could you do that sooner rather than later?

We don't really have the types, since they're both `String` :-/

Design consistency absolutely is a goal, whether specific things matter is up to our product and design people though, not us (although the technical constraints (like us not supporting HTML) are obviously totally relevant here).

adamnfish: @andrewcb can you follow up these points with @janua and our PM / design

andrewcb: What is the original kickerText function used for elsewhere? I did a `grep kickerText **/*.scala` in `frontend` and it yielded no results.

janua: The original problem here was to get the text "Live" for `LiveKicker`, and "Breaking" for `BreakingNewsKicker`, and so on. You do have the types for this, and this is resolved for you from `Content`.

What the editors enter for a custom kicker is always going to be the type `String`. And, you have the appropriate types for that too.

So are we talking about converting the types to some text representation, or talking about what exists inside `FreeHtmlKicker`?

andrewcb:
What I'm looking for is something that will return the canonical text for any `ItemKicker`. I.e., “Breaking” for `BreakingNewsKicker`, the tag name for `TagKicker`, and so on. We need plain text, so HTML kickers, if they cannot be made into plain text, should be returned as `None`.

fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
andrewcb: What I'm looking for is something that will return the canonical text for any `ItemKicker`. I.e., “Breaking” for `BreakingNewsKicker`, the tag name for `TagKicker`, and so on. We need plain text, so HTML kickers, if they cannot be made into plain text, should be returned as `None`.

andrewcb: Another option could be to add a `asPlainText` method to the `ItemKicker` trait. Would you be averse to this?

adamnfish: What have we said about mixing model definitions with logic...

andrewcb: It's hardly logic if it takes no parameters.

andrewcb: In any case, the `asPlainText` would return either a constant string (i.e., `Breaking`) or one of the kicker's data fields, unmodified.  This would assist in keeping the canonical names of the kickers in one place, so they don't diverge across different projects.

andrewcb: Also, should the existing `kickerText` function (to be renamed `kickerHTML`) not return `Some(...)` for the data-less kickers like `BreakingNewsKicker`? The plain-text function I want would do this; would keeping the behaviours of the two functions different be a good idea, or would there be a case for making them consistent?

andrewcb: If the answer to the last one is “no, let's leave it as is”, then perhaps `kickerHTML` is not the best name for it; `kickerContents` may be a better name (it returns the contents of kickers which have them, in text or HTML, and `None` for contentless kickers).

adamnfish:
This doesn't have any tests at the moment. There's not much here, but it'd definitely be good to put something in to catch regressions that would upset the apps. For example, testing that the HTML-style ones return None in kicker text would be valuable.

adamnfish:
:+1: from me, up to you though @janua

janua:
:+1:
I will merge this when it is ready to go out.

PR #74, opened 2015-02-25
Fix requesting without Snaps

janua:
This fixes requesting with `FAPI` via `collectionContentWithoutSnaps`.

Snaps were still getting through, and the test was wrong.

@adamnfish @robertberry


robertberry:
:+1:

PR #76, opened 2015-03-03
kickerText(FreeHtmlKicker* ) now returns text of plain-text kickers

andrewcb:
This is a workaround for the current tendency to use `FreeHtmlKicker` for plain-text keyword kickers; once there is a `FreeTextKicker` or similar, it will hopefully be obsolete.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
@@ -91,7 +91,7 @@ object ItemKicker {
   }
 
   def stringIfPlainText(string: String) : Option[String] = {
-    val StringWithHtml = "<\\w+>".r.unanchored
+    val StringWithHtml = "<\\w+.*>".r.unanchored
     string match {
       case StringWithHtml() => None
       case _ => Some(string)
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
@@ -90,14 +90,22 @@ object ItemKicker {
     case _ => None
   }
 
+  def stringIfPlainText(string: String) : Option[String] = {
+    val StringWithHtml = "<\\w+>".r.unanchored
+    string match {
+      case StringWithHtml() => None
+      case _ => Some(string)
+    }
+  }
+
   /** Return a plain-text representation of a kicker */
   def kickerText(itemKicker: ItemKicker): Option[String] = itemKicker match {
     case BreakingNewsKicker => Some("Breaking")
     case AnalysisKicker => Some("Analysis")
     case ReviewKicker => Some("Review")
     case CartoonKicker => Some("Cartoon")
-    case FreeHtmlKicker(_) => None
-    case FreeHtmlKickerWithLink(_, _) => None
+    case FreeHtmlKicker(text) => stringIfPlainText(text)
+    case FreeHtmlKickerWithLink(text, _) => stringIfPlainText(text)
     case _ => kickerContents(itemKicker)
   }
 }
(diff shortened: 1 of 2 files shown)


fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
```diff
@@ -90,14 +90,22 @@ object ItemKicker {
     case _ => None
   }
 
+  def stringIfPlainText(string: String) : Option[String] = {
+    val StringWithHtml = "<\\w+>".r.unanchored
```
adamnfish: This regex is inadequate but I think the approach is flawed - using regular expressions to parse HTML is a minefield. What about a test case like `<b class="tmp">Something</b>`?

When you need to parse HTML in strings, generally first consider an approach using `scala.xml.XML.loadString`.

andrewcb: Good catch about the regex.  
The problem with `scala.xml.XML.loadString` is that it does not determine whether something contains HTML tags, but rather whether it is well-formed XML (a much narrower specification). This is not particularly useful for this task.

adamnfish: Consider

    scala.xml.XML.loadString(string).text == string

andrewcb: scala.xml.XML.loadString("foo")
org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content is not allowed in prolog.

scala> scala.xml.XML.loadString("<p> foo <br> bar </p>")
org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 20; The element type "br" must be terminated by the matching end-tag "</br>".

janua:
I don't like this as a regex. What about `jsoup`? Or, anything that's not regex.

andrewcb:
Wouldn't adding a HTML parser (not part of the library right now) be overengineering for the problem (determining whether or not a string contains HTML tags)? Detecting HTML (as opposed to parsing HTML) is IMHO not a problem that's beyond regular expressions.

janua:
I'd rather that than regex abuse.

This reminds me of http://stackoverflow.com/a/1732454

andrewcb:
If we were talking about parsing HTML, I'd agree with you wholeheartedly. Attempting to parse HTML with regular expressions is asking for trouble. However, I believe that this use-case is one for which regular expressions would suffice, and indeed, a full parser would be overkill, for the following reasons:
* We are not parsing HTML, but _detecting_ its presence/absence. The nature/content of the tags, if found, is irrelevant.
* The problem of determining whether or not an input contains a HTML tag can be encoded by a regular expression. One could specify possible tags precisely in a somewhat long expression with attributes in brackets), but in this case (matching against content coming from Facia, and determining whether there are any tags in it), looking for <something...> is probably sufficient. 
* The regular expression included would not get false negatives (i.e., there is no way for HTML tags to slip through). It's tight enough to not give false positive for possible phrases such as "2>1" or "I <3 textspeak", which might conceivably turn up in kickers, though these would be edge-cases. In the worst case, if there's a kicker like, say, "I <3 cats>dogs", the failure case is that the kicker is not shown. Which is less than ideal though not catastrophic; in any case, if the Free HTML Kicker is going to be replaced by a Free Text Kicker, this solution is only a bandaid until this happens, and in any case yields a better result than dropping Free HTML Kickers in all cases (as we do now).

Though, if `FreeTextKicker`s are coming, perhaps the thing to do is to do nothing, make no attempt to determine whether a string contains HTML tags or not, and live with `FreeHTMLKicker`s not being shown on MAPI for now. Certainly, if this is the case, adding an entire HTML parser library to produce a boolean in one function strikes me as overkill, unless there's a compelling case for it.

adamnfish:
I agree with Andrew on this one although it'd be nice to have a couple of (or at least one?) real kicker examples in as tests rather than just the basic `<tag>text</tag>` vs `text` that's there now?

janua:
@adamnfish @andrewcb Yes, I am looking into this now. How urgent is it that this gets merged now?

I can give you some real world kicker examples very soon as @piuccio is writing the traverse of collections now.

adamnfish:
@janua super urgent. Don't worry about the real data for now, there are a couple more examples in there so the tests look a bit better.

janua:
:+1:

PR #77, opened 2015-03-04
Change logic around itemKicker for SupportingCuratedContent

janua:
In the past, the `config` was ignored when resolving the `ItemKicker` for a `SupportingCuratedContent` by passing a `None`.

This puts that behaviour back in.

@adamnfish @robertberry


robertberry:
:+1:

PR #78, opened 2015-03-06
Use ResolvedMetaData when creating ImageCutout

janua:
Depending on the type of content, `imageCutoutReplace` should default to `true` when not set.

@robertberry @adamnfish


robertberry:
:+1:

PR #79, opened 2015-03-09
Add getters to FaciaContent

DiegoVazquezNanini:
I added a bunch of getters to the FaciaContent  to avoid using the trait and to Allow Francis to refactor this file in the future.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
@@ -57,150 +57,149 @@ sealed trait FaciaContent
 
 object FaciaContent {
 
-  implicit class FaciaContentHelper(faciaContent: FaciaContent) {
+  def fold[T](fc: FaciaContent)(c: (CuratedContent) => T, scc: (SupportingCuratedContent) => T,
+              ls: (LinkSnap) => T, las: (LatestSnap) => T): T = fc match {
+    case curatedContent: CuratedContent => c(curatedContent)
+    case supportingCuratedContent: SupportingCuratedContent => scc(supportingCuratedContent)
+    case linkSnap: LinkSnap => ls(linkSnap)
+    case latestSnap: LatestSnap => las(latestSnap)
+  }
 
-    def fold[T](c: (CuratedContent) => T, scc: (SupportingCuratedContent) => T,
-                ls: (LinkSnap) => T, las: (LatestSnap) => T): T = faciaContent match {
-      case curatedContent: CuratedContent => c(curatedContent)
-      case supportingCuratedContent: SupportingCuratedContent => scc(supportingCuratedContent)
-      case linkSnap: LinkSnap => ls(linkSnap)
-      case latestSnap: LatestSnap => las(latestSnap)}
+  def id(fc: FaciaContent): String = fold(fc)(
+    curatedContent => curatedContent.content.id,
+    supportingCuratedContent => supportingCuratedContent.content.id,
+    linkSnap => linkSnap.id,
+    latestSnap => latestSnap.id)
 
-    def id: String = fold(
-      curatedContent => curatedContent.content.id,
-      supportingCuratedContent => supportingCuratedContent.content.id,
-      linkSnap => linkSnap.id,
-      latestSnap => latestSnap.id)
+  def content(fc: FaciaContent): Option[Content] = fold(fc)(
+    curatedContent => Option(curatedContent.content),
+    supportingCuratedContent => Option(supportingCuratedContent.content),
+    linkSnap => None,
+    latestSnap => latestSnap.latestContent)
 
-    def content: Option[Content] = fold(
-      curatedContent => Option(curatedContent.content),
-      supportingCuratedContent => Option(supportingCuratedContent.content),
-      linkSnap => None,
-      latestSnap => latestSnap.latestContent)
+  def maybeContent(fc: FaciaContent) = content(fc)
 
-    def maybeContent = content
+  def supportingContent(fc: FaciaContent): List[FaciaContent] = fold(fc)(
+    curatedContent => curatedContent.supportingContent,
+    supportingCuratedContent => Nil,
+    linkSnap => Nil,
+    latestSnap => Nil)
 
-    def supportingContent: List[FaciaContent] = fold(
-      curatedContent => curatedContent.supportingContent,
-      supportingCuratedContent => Nil,
-      linkSnap => Nil,
-      latestSnap => Nil)
+  def headline(fc: FaciaContent): String = fold(fc)(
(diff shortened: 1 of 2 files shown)


adamnfish:
There are no tests!

We don't need to test every detail, but stuff where there's logic si worth testing (e.g. falling back to `"Missing Headline"` - since that String is hard coded clients might implement against it and the test will stop changes breaking things)

fapi-client/src/test/scala/com/gu/facia/api/models/FaciaContentHelperTest.scala
```diff
@@ -0,0 +1,36 @@
...
+    val snap = LatestSnap("myId", None, None, None, None, None, None, "myGroup", None, false, false, false, false, false, false, None, false, None, None, false, false)
+    FaciaContent.headline(snap) should equal("Missing Headline")
+  }
+
+  "should return the headline for a CuratedContent" in {
+    val cc = CuratedContent(null, Nil, "The headline", None, None, "myGroup", None, false, false, false, false, false, None, false, None, None, false, false)
+    FaciaContent.headline(cc) should equal("The headline")
+  }
+
+  //href
```
adamnfish: what's this comment here for?

fapi-client/src/test/scala/com/gu/facia/api/models/FaciaContentHelperTest.scala
```diff
@@ -0,0 +1,36 @@
+package com.gu.facia.api.models
+
+import com.gu.contentapi.client.model.Content
+import org.scalatest.{FreeSpec, Matchers}
+
+class FaciaContentHelperTest extends FreeSpec with Matchers {
+
+  //headline
```
adamnfish: what's this comment here for?

adamnfish:
Nice to have those tests, also interesting that adding tests exposed that the other way of implementing the methods was less testable!

Not sure why there are a few `null`s in the tests though? Is there not a sensible default value? If you find yourself using nulls in tests to get around having to make concrete instances of things then consider using mocks. They're nulls as well, but without all the boiler plate. In these tests mocks might make it much clearer because you'd be doing the setup the test needs explicitly, rather than having the important fields hidden in with all the default values (`false`, `None`, etc). Up to you!

DiegoVazquezNanini:
@adamnfish the last commit address all your comment, except the suggestion of using mocks.

adamnfish:
Looks fine to me, but be nice to get a thumbs up from the fronts team because I can't speak for the correctness of that logic :-/

DiegoVazquezNanini:
@johnduffell @robertberry any comments ?

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -55,6 +55,153 @@ object ImageCutout {
...
+    linkSnap => linkSnap.id,
+    latestSnap => latestSnap.id)
+
+  def content(fc: FaciaContent): Option[Content] = fold(fc)(
+    curatedContent => Option(curatedContent.content),
+    supportingCuratedContent => Option(supportingCuratedContent.content),
+    linkSnap => None,
+    latestSnap => latestSnap.latestContent)
+
+  def maybeContent(fc: FaciaContent) = content(fc)
```
robertberry: Why this alias?

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -55,6 +55,153 @@ object ImageCutout {
...
+  def supportingContent(fc: FaciaContent): List[FaciaContent] = fold(fc)(
+    curatedContent => curatedContent.supportingContent,
+    supportingCuratedContent => Nil,
+    linkSnap => Nil,
+    latestSnap => Nil)
+
+  def headline(fc: FaciaContent): String = fold(fc)(
+    curatedContent => curatedContent.headline,
+    supportingCuratedContent => supportingCuratedContent.headline,
+    linkSnap => linkSnap.headline.getOrElse("Missing Headline"),
```
robertberry: Why not an Option?

(I'd tend to push rendering something else to the most outer layer possible to give consumers of the client the choice.)

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -55,6 +55,153 @@ object ImageCutout {
...
+    latestSnap => latestSnap.headline.orElse(latestSnap.latestContent.map(_.webTitle)).getOrElse("Missing Headline"))
+
+  def href(fc: FaciaContent): String = fold(fc)(
+    curatedContent => curatedContent.href.getOrElse("Missing href"),
+    supportingCuratedContent => supportingCuratedContent.href.getOrElse("Missing href"),
```
robertberry: Ditto to above.

robertberry:
Neat. I like the fold. :+1:

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
DiegoVazquezNanini: @janua likes the maybe* for Options, but the actual name of the attribute is content so I guess I add that alias to make @janua happy.

robertberry: *shrugs* Seems kind of redundant when it's in the type signature (which always deserves more attention than the name!) but I'm not bothered if it's a standard that's already been set in this project.

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
DiegoVazquezNanini: Yes, that make a lot of sense. Let me fix it.

DiegoVazquezNanini:
Thansk @robertberry and @adamnfish

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -55,6 +55,151 @@ object ImageCutout {
 
 sealed trait FaciaContent
 
+object FaciaContent {
+
+  def fold[T](fc: FaciaContent)(c: (CuratedContent) => T, scc: (SupportingCuratedContent) => T,
+              ls: (LinkSnap) => T, las: (LatestSnap) => T): T = fc match {
```
johnduffell: I know it's already merged, but I'd recommend giving things more descriptive names where possible.  We usually have autocomplete anyway, and it helps my poor brain =)

DiegoVazquezNanini:
That is a fair comment. I will follow your advice in the future. Thanks @johnduffell :)

PR #80, opened 2015-03-12
Add canonical flag

robertberry:
CC @adamnfish @DiegoVazquezNanini


janua:
Canonical is meant to be at a `path` level, not at `collection` level, with a fallback if the collection doesn't exist for the path.

adamnfish:
@janua yeah, cheers Francis. I already spoke with about this @robertberry IRL which is why this didn't get merged or anything.

adamnfish:
Replaced by #84

PR #82, opened 2015-03-16
Sorry I need to do this inside MAPI instead

DiegoVazquezNanini:
Hi @robertberry,
Today I realize that only MAPI needs this field and therefore it shouldn't be polluting the library.
No rush, but we should merge this PR before the next release of the facia-client.

Cheers


adamnfish:
:+1:

janua:
:+1:

Issue #83, opened 2015-03-17
Non empty lists


adamnfish:
It'd be nice to try and embed more of the business rules in the types to make implementations simpler. In particular, it's impossible for there to be 0 collections on a front, so that could be represented with a non-empty-list type. This would allow us to safely call `head` to get the first collection without having to provide dummy fallback content (see the upcoming feature on "canonical collections").

Scalaz has a `NonEmptyList` type that would suit this very well.

Not urgent, but it'd make the code clearer and adding some new features easier!

PR #84, opened 2015-03-17
Canonical collections

adamnfish:
This PR replaces https://github.com/guardian/facia-scala-client/pull/80. The canonical container either comes from the FrontJson, or is the first container on the front.

Note that this PR contains two commits, the second specifies the edition network fronts' canonical containers as well. While we wait for the tools to support this feature, the network front canonical containers can be hard-coded. This commit can be rolled back when the tools support is in place and the correct JSON is actually coming through.

Don't merge this yet, it's open for comment/review, I'm going to test it by integrating it with MAPI.

Note that I'm assured there can never be a front with 0 containers. I've raised a an issue (https://github.com/guardian/facia-scala-client/issues/83) suggesting we capture this by making that a `NonEmptyList`. That would remove the need for the `"no collections"` fallback string.

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -76,7 +76,8 @@ case class FrontJson(
   imageHeight: Option[Int],
   isImageDisplayed: Option[Boolean],
   priority: Option[String],
-  isHidden: Option[Boolean]
+  isHidden: Option[Boolean],
+  canonical: Option[String]
 )
 
 object ConfigJson {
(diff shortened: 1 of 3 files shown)


adamnfish:
Question: What happens if the editorially chosen canonical collection isn't actually on the front? This shouldn't be possible at the tools level, but is a totally reasonable edge case when caching etc is taken into account.

Should it silently fall back to the top container? Leave the front with no canonical container (which will break things downstream)? Throw an exception?

Thoughts @janua @stephanfowler

adamnfish:
(p.s. this problem is happens at the moment on CODE because the `uk`, `us`, `au` paths don't contain the hard-coded collection IDs, annoyingly. Clearly that's a bug in my hard-coded "waiting for tools" workaround but the question above is relevant to the solution to this bug.

janua:
@adamnfish I think we should fallback, as that is the current behaviour; a reliable fallback consistent in code.
I think we should do both. Make it `Option` but also have a method of getting the solid underlying type without the `Option`.

janua:
:+1:

PR #87, opened 2015-04-07
Request treats

janua:
This adds the ability to request `treats` for a given `Collection`.

For normal `items`/`trails`, it requests via a `SearchQuery`. For dream snaps, it requests via an `ItemQuery`.

@adamnfish @robertberry

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -87,6 +87,19 @@ object FAPI {
     for(snapContent <- ContentApi.latestContentFromLatestSnaps(capiClient, allSnaps, adjustItemQuery))
       yield snapContent}
 
+  def getTreatsForCollection(collection: Collection, adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
+                      (implicit capiClient: GuardianContentClient, ec: ExecutionContext) = {
+    val treatsRequest = Collection.treatsRequestFor(collection)
+    ContentApi.buildHydrateQueries(capiClient, treatsRequest.treatIds, adjustSearchQuery) match {
+      case Success(hydrateQueries) =>
+        for {
+          hydrateResponses <- ContentApi.getHydrateResponse(capiClient, hydrateQueries)
+          snapContent <- ContentApi.latestContentFromLatestSnaps(capiClient, treatsRequest.latestSnapsRequest, adjustItemQuery)
+          setOfContent = ContentApi.itemsFromSearchResponses(hydrateResponses)}
+    yield Collection.treatContent(collection, setOfContent, snapContent)
+      case Failure(error) =>
+        Response.Left(UrlConstructError(error.getMessage, Some(error)))}}
+
   def collectionContentWithoutSnaps(collection: Collection, adjustSearchQuery: AdjustSearchQuery = identity)
                                 (implicit capiClient: GuardianContentClient, ec: ExecutionContext): Response[List[FaciaContent]] = {
     val collectionWithoutSnaps = Collection.withoutSnaps(collection)
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -9,7 +9,6 @@ import com.gu.facia.client.ApiClient
 import com.gu.facia.client.models.TrailMetaData
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
 
 
 object FAPI {
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -89,12 +89,12 @@ object FAPI {
 
   def getTreatsForCollection(collection: Collection, adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
                             (implicit capiClient: GuardianContentClient, ec: ExecutionContext) = {
-    val treatsRequest = Collection.treatsRequestFor(collection)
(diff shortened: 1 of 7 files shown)


fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -87,6 +87,19 @@ object FAPI {
     for(snapContent <- ContentApi.latestContentFromLatestSnaps(capiClient, allSnaps, adjustItemQuery))
       yield snapContent}
 
+  def getTreatsForCollection(collection: Collection, adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
```
adamnfish: The formatting's a bit all over the place in a couple of places here. The arg list and the for comp jump out?

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -83,7 +98,21 @@ object Collection {
       collection.live
       .filter(_.isSnap)
       .filter(_.safeMeta.snapType == Some("latest"))
-      .flatMap(snap => snap.safeMeta.snapUri.map(uri => snap.id -> uri)).toMap)
+      .flatMap(snap => snap.safeMeta.snapUri.map(uri => snap.id -> uri))
+      .toMap)
+
+  def treatsRequestFor(collection: Collection): TreatsRequest = {
```
adamnfish: I don't much like this function. It's only used in one place, so why not inline it somewhere? Even returning a tuple might make it a bit simpler. This is wrapped up with my comment above about the logic in the FAPI method though, so maybe it needs a little thought up there and that'll make all this stuff simpler as well.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -87,6 +87,19 @@ object FAPI {
...
+                      (implicit capiClient: GuardianContentClient, ec: ExecutionContext) = {
+    val treatsRequest = Collection.treatsRequestFor(collection)
+    ContentApi.buildHydrateQueries(capiClient, treatsRequest.treatIds, adjustSearchQuery) match {
+      case Success(hydrateQueries) =>
```
adamnfish: It seems like there's a bit more logic here than is ideal? The goal is that the FAPI class contains methods that do nothing but trivially thread together Responses in a for-comp. Here we've got nested logic, matching and explicit error handling. This is connected to the fact that the `treatsRequestFor` method is a bit odd as it stands. I think this collectively this is all just 'code smell' that stems from this function trying to do too much stuff.

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
janua: I don't agree with in-lining this or with the `tuple` suggestion. This is purposely returning a `TreatsRequest` to deal with to two different methods of requesting items. Where were you suggesting to in-line this?

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
janua: I see what you mean, but this is the way `buildHydrateQueries` works, and this PR is not about refactoring that. If this format isn't good enough for `getTreatsForCollection`, when why is it good enough for `getContentForCollection`?
Perhaps the solution is to change the signature to `buildHydrateQueries` to return a `Response[Seq[SearchQuery]]` instead of a `Try[Seq[SearchQuery]]`.

janua:
That commit hid your comment!! Unintentional, promise :smile:

janua:
@adamnfish Changed my mind, I did it in this PR. `buildHydrateQueries` now returns a `Response`. And your `for` comprehensions are clean.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
adamnfish: Formatting's still a bit Pete Tong here...

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -17,22 +17,18 @@ object ContentApi {
...
-  def buildHydrateQueries(client: ContentApiClientLogic, ids: List[String], adjustSearchQuery: AdjustSearchQuery = identity): Try[Seq[SearchQuery]] = {
+  def buildHydrateQueries(client: ContentApiClientLogic, ids: List[String], adjustSearchQuery: AdjustSearchQuery = identity): Response[Seq[SearchQuery]] = {
     def queryForIds(ids: Seq[String]) = adjustSearchQuery(client.search
       .ids(ids mkString ",")
       .pageSize(ids.size)
       .showFields("internalContentCode"))
 
-    Try {
-      IdsSearchQueries.makeBatches(ids)(ids => client.getUrl(queryForIds(ids))) match {
+    IdsSearchQueries.makeBatches(ids)(ids => client.getUrl(queryForIds(ids))) match {
```
adamnfish: Are you sure the `Try` wasn't being used to catch exceptions in the block?

adamnfish:
There are a few places where the indentation is all weird in the code and tests. A few o fthe for comps have braces in weird places and funny indenting. Looks ok otherwise, much better now we're using response in place of `Try` (assuming it is safe).

:+1: if the Try/Response is safe

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
janua: Yeah, was thinking that myself. I'm not sure, I don't think to. @robertberry is a fan of `Exceptions`. What do you think @robertberry ?

robertberry: Not sure why I was using `Try` there

robertberry: It's probably to do with URL construction, which can throw exceptions.

PR #88, opened 2015-04-09
Fix kickers

janua:
When `LatestSnaps` were being created, they were using `ResolvedMetaData.Default` as the `ResolvedMetaData` instead of resolving from the `Trail` `MetaDataCommonFields`.

This also passes along an `Option[Content]` to the kicker resolving for `LatestSnap`.

@adamnfish @robertberry

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
@@ -41,7 +41,10 @@ object ItemKicker {
     }
   }
 
-  def fromTrailMetaData(trailMeta: MetaDataCommonFields): Option[ItemKicker] = fromContentAndTrail(None, trailMeta, ResolvedMetaData.Default, None)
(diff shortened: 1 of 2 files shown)


fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
```diff
@@ -41,7 +41,10 @@ object ItemKicker {
     }
   }
 
-  def fromTrailMetaData(trailMeta: MetaDataCommonFields): Option[ItemKicker] = fromContentAndTrail(None, trailMeta, ResolvedMetaData.Default, None)
+  def fromTrailMetaData(trailMeta: MetaDataCommonFields): Option[ItemKicker] = fromContentAndTrail(None, trailMeta, ResolvedMetaData.fromTrailMetaData(trailMeta), None)
```
adamnfish: Are these functions actually useful? Why can't the rest of the code just call the `fromContentAndTrail` function with the right arguments?

janua: Of course they could do that. I think that makes the code a lot messier. And this is a common enough format to not be unfamiliar.

adamnfish:
:+1:

PR #90, opened 2015-04-10
Carry embed values

janua:
In `facia-tool`, we have three fields that are used extensively with `Snap`s: `snapType`, `snapUri` and `snapCss`. They are used for embeds, such as when a football widget gets embedded on a page via a `LinkSnap`.

`Thrashers` on the other hand, use the same fields but are also an actual piece of content (For when they don't get upgraded). So for `thrashers` to work, normal content (`CuratedContent`) needs to carry these `snapX` fields, except now that it's no longer a `Snap` it makes no sense to call them `snapX`, they will now be known as `embedX`.

`SupportingCuratedContent` does not carry the fields as I don't think they need to right now.

@adamnfish @robertberry


adamnfish:
:+1:

PR #92, opened 2015-04-14
Change ImageCutout fromContentTags

janua:
This changes `ImageCutout.fromContentTags` so return an `ImageCutout` when there is only one contributor in the `tags` for the `Content`.

This brings back the original behaviour which got lost in the transfer from `frontend`.


robertberry:
:+1:

PR #93, opened 2015-04-14
Use resolved metadata for LatestSnap

janua:
This uses `ResolvedMetaData` for the resolution of the properties for `LatestSnap`. This brings it in line with what happens for a `CuratedContent` and a `SupportingCuratedContent`, meaning that defaults will now appear such as if the `LatestSnap` is a `comment` piece or a `video`.

@robertberry @DiegoVazquezNanini


robertberry:
:+1:

PR #94, opened 2015-04-15
Type helpers for Content and FaciaContent

janua:
This move and adds some helpers for the types `Content` and `FaciaContent`.

I had been using a more comprehensive list of methods which used the `fold` from `FaciaContent` in `frontend` up until now, and I have just copied this over for use here.

I have also added helpers for the type `Content`.

This paves the way for `CardStyle`.

@robertberry @adamnfish @DiegoVazquezNanini


adamnfish:
Tests?

fapi-client/src/test/scala/com/gu/facia/api/utils/FaciaContentUtilsTest.scala
```diff
@@ -0,0 +1,357 @@
...
+      fields = Option(fields),
+      Nil, None, Nil, None)
+
+    "should return a false for a LinkSnap" in {
+      val linkSnap = makeLinkSnap("some-link-snap")
+      FaciaContentUtils.isLive(linkSnap) should be (false)
+    }
+
+    "should return true and false on inner content for LatestSnap" in {
+      val latestSnap = makeLatestSnap("some-latest-snap", contentWithFields(Map("liveBloggingNow" -> "true")))
```
adamnfish: There are a few like this here. Should these not be multiple tests?

fapi-client/src/test/scala/com/gu/facia/api/utils/FaciaContentUtilsTest.scala
```diff
@@ -0,0 +1,357 @@
+package com.gu.facia.api.utils
+
+import com.gu.contentapi.client.model.{Tag, Content}
+import com.gu.facia.api.models.{SupportingCuratedContent, CuratedContent, LatestSnap, LinkSnap}
+import org.joda.time.DateTime
+import org.scalatest.{FreeSpec, Matchers}
+
+class FaciaContentUtilsTest extends FreeSpec with Matchers {
+
+  def makeLinkSnap(linkSnapId: String) = LinkSnap(
```
adamnfish: This is a *lot* of setup. Worth thinking about whether this can be pulled out into fixtures for the tests that could then be shared?

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentImplicits.scala
```diff
@@ -0,0 +1,119 @@
...
+
+import com.gu.contentapi.client.model.{Element, Tag, Content}
+import com.gu.facia.api.models.{ImageReplace, ImageCutout, FaciaContent}
+import org.joda.time.DateTime
+
+import scala.util.Try
+
+object FaciaContentImplicits {
+
+  implicit class FaciaContentImplicit(fc: FaciaContent) {
```
adamnfish: Do we really need both ways of accessing these? Is it not better to encourage users of this library to pick one way or the other? It means there's the burden of keeping these in sync forever more and precisely one of the approaches is redundant.

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentImplicits.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    def trailText: Option[String] = FaciaContentUtils.trailText(fc)
+
+    def maybeWebTitle: Option[String] = FaciaContentUtils.maybeWebTitle(fc)
+
+    def webTitle: String = FaciaContentUtils.webTitle(fc)
+
+    def linkText = FaciaContentUtils.linkText(fc)
+
+    def imageReplace(fc: FaciaContent): Option[ImageReplace] = FaciaContentUtils.imageReplace(fc)
```
adamnfish: I don't think these should be taking `FaciaContent` as an arg?

janua: No, they shouldn't.

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentImplicits.scala
janua: I use the implicits, it's much cleaner to use. I can keep this in frontend only if you prefer.

If I had to pick, I would pick the implicit way for `FaciaContent`. I have only provided the other way because you said you don't like it @adamnfish.

fapi-client/src/test/scala/com/gu/facia/api/utils/FaciaContentUtilsTest.scala
janua: Maybe. What is testing really? Strictly one assert per test case or grouping another way?

fapi-client/src/test/scala/com/gu/facia/api/utils/FaciaContentUtilsTest.scala
janua: True true

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentImplicits.scala
janua: I have fixed this upstream in #95

fapi-client/src/test/scala/com/gu/facia/api/utils/FaciaContentUtilsTest.scala
adamnfish: Not "strictly" but definitely as a strongly-encouraged rule of thumb. This has three entirely distinct and unconnected parts, each with their own assert. As I said in the next PR, using FreeSpec you can always group them together in an enclosing block if that makes it clearer / saves test message duplication.

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentImplicits.scala
adamnfish: I don't understand why you'd have both, it's just a burden. Part of good library design is choosing a good usage and then encouraging it. Having multiple ways of doing the same thing is confusing.

The pure way is certainly more flexible and removes the indirection of using implicits, so as you say I think that's the better approach and it would get my vote for the one to keep.

adamnfish: (what would you do if this was a Haskell application?)

janua: @adamnfish Haha, good question. The record syntax makes this a bit less painful, and you have no `implicits` so the only way is the pure way; you don't really have a choice. At least in Haskell it would be slightly more concise in that you don't have to use some object with methods on it to access a method to pass something in, you just have the property sort of floating around for the type, kind of like implicits but in reverse syntax and without the dot.

I don't think this is multiple ways of doing the same thing at all; it's just a different way to use the underlying methods. This is how I'm using it everyone in `frontend`.

I don't personally don't like named imports, such as `import FaciaContentUtils.isLive`, to use `isLive(faciaContent)`, and I don't like having to use the fully qualified path for the method everywhere either `FaciaContentUtils.isLive(faciaContent)`.

adamnfish: If you just do `import FaciaContentUtils._` wherever you want to use them it's the same as Haskell...

adamnfish: You'd have to do an import to get the implicit too, remember.

This is absolutely two ways of doing the same thing! I strongly believe you should pick one (and only slightly less strongly believe it should be the pure version you keep). The errors in this PR are an example of why it's bad, and maintaingin both is going to be SUCH A DRAG. Imagine every time someone in the mobile team adds one that they need and every time you have to be like "oh yeah, you need to update it here as well" and they say "what?! Why are you doing the the same thing in two places this is mad?!" and you say "I don't think this is multiple ways of doing the same thing at all; it's just a different way to use the underlying methods. This is how I'm using it everyone in frontend." Obvs up to you though.

janua: All our masses of users of the lib are going to get so confused we will need a wiki!
Oh i'll just move it back to `frontend` seeing as it's such a big problem.

PR #95, opened 2015-04-21
Add CardStyle

janua:
This adds `CardStyle` to `facia-scala-client`, and exposes it on `CuratedContent`, `SupportingCuratedContent` and `LatestSnap`. For `LinkSnap`, it is `Default`.

I had to create a `ContentProperties` due to the limit on values on case classes; this groups all the `Booleans` together.

`ExternalLinks` is also new, I am not sure if it belongs here.

This depends on #94, and it a PR on top of that. So does not include the changes there.


fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
```diff
@@ -0,0 +1,101 @@
+package com.gu.facia.api.utils
+
+import com.gu.contentapi.client.model.Content
+import com.gu.facia.api.utils.ContentApiUtils._
+import com.gu.facia.client.models.MetaDataCommonFields
+
+object CardStyle {
+
+  def isExternalLink(content: Content, metaDataFields: MetaDataCommonFields): Boolean = (for {
```
adamnfish: This function is only used in one place and has identical arguments to the function that calls it. It's not tested or anything so does it need to be here separately?

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
```diff
@@ -0,0 +1,101 @@
...
+    snapType <- metaDataFields.snapType
+    href <- metaDataFields.href.orElse(content.safeFields.get("href"))
+  } yield snapType == "link" && ExternalLinks.external(href)) getOrElse false
+
+  def apply(content: Content, trailMetaData: MetaDataCommonFields): CardStyle = {
+    if (isExternalLink(content, trailMetaData)) {
+      ExternalLink
+    } else if (content.tags.exists(_.id == "news/series/hsbc-files")) {
```
adamnfish: Whaaa?!

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
```diff
@@ -0,0 +1,101 @@
...
+      Letters
+    } else if (content.isFeature) {
+      Feature
+    } else {
+      Default
+    }
+  }
+}
+
+sealed trait CardStyle {
```
adamnfish: Would this be simpler as just a string?

  object CardStyles {
    val specialReport = "special-report"
    val toneString = "live"
    // ... etc ...
  }

Even if not, you can take out the explicit type annotations for each of these. Also, why are they defs instead of vals? Using def for the abstract member is good (so it can be implemented either way) but the implementations don't need to be lazy or recalculated every time.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
```diff
@@ -0,0 +1,32 @@
+package com.gu.facia.api.utils
+
+import java.net.URI
+
+import scala.util.Try
+
+object ExternalLinks {
+  val corsOrigins = List(
```
adamnfish: Why are they called this when it has nothing to do with CORS? It just serves to confuse the purpose here, which is to work out if a link is an internal Guardian one. Even inlining this list in the `GuardianDomains` property would be clearer here, I think.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
```diff
@@ -0,0 +1,32 @@
...
+    "http://www.theguardian.com",
+    "https://www.theguardian.com",
+    "https://profile.theguardian.com",
+    "https://composer.gutools.co.uk",
+    "https://composer.release.dev-gutools.co.uk",
+    "https://composer.code.dev-gutools.co.uk",
+    "http://preview.gutools.co.uk",
+    "https://preview.gutools.co.uk")
+
+  val GuardianDomains = corsOrigins flatMap { uri =>
```
adamnfish: Why's this a capital `G`? This may be complex enough to benefit from a type annotation.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
```diff
@@ -0,0 +1,32 @@
...
+    Try {
+      new URI(uri).getHost.stripPrefix("www.")
+    }.toOption
+  }
+
+  def external(url: String) = Try(Option(new URI(url).getHost).exists({ host => !GuardianDomains.exists({ domain =>
```
adamnfish: This is a very long line!

Might also be clearer with a type annotation (or if it is renamed to `isExternal` it'll be clearer this is just a test returning a Bool? (with all the `Try` indirection it isn't at all clear from first glance what this does)

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
```diff
@@ -0,0 +1,32 @@
...
+    host == domain || host.endsWith(s".$domain")
+  })})).getOrElse(false)
+
+  def internalPath(url: String) = if (external(url)) None else Try {
```
adamnfish: I'd quite like a type annotation here as well? Again, there are trys and options involved in non-obvious ways.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
```diff
@@ -0,0 +1,32 @@
+package com.gu.facia.api.utils
+
+import java.net.URI
+
+import scala.util.Try
+
+object ExternalLinks {
```
adamnfish: These methods don't have any tests?

adamnfish:
There don't seem to be any tests for the new stuff?

This codebase is starting to get really complicated now - there are a lot of obscure bits of logic (that hsbc-files special case?!). One side effect is that the types are starting to become quite monolithic. Keep in mind whether it's possible to split them up into separate bits of functionality, rather than adding more and more properties onto one giant type, lest this end up worse than the CAPI client... Not saying that's what needs to be done here, but generally can there be a function that gets the styles for a card, rather than adding it (and every conceivable other property) onto the one type of doom.

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
janua: @adamnfish It has to exist somewhere unfortunately! And if we want to share this with apps, it has to be in this library!
Otherwise, we could let app not have `SpecialReports`, which is I doubt what people want!!
It's nasty.

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
janua: T == T

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
janua: True true true. No idea why it was made this way. Ask ORB. @robertberry

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
janua: They do, I didn't port them. I shall.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
janua: Bad name, they are actually the `CORS` list from `frontend`, but in this context they make no sense.

fapi-client/src/main/scala/com/gu/facia/api/utils/ExternalLinks.scala
janua: Now added

janua:
@adamnfish Yes, I completely agree. I think that could be a much better approach to making the types a bit smaller, just have methods for getting things instead of resolving them up front. This sort of thing is similar to our up-front/not-up-front argument :wink:

adamnfish:
Looks much better now! Nice one :+1:

PR #96, opened 2015-04-21
Draft collections

janua:
This allows one to use `FAPI` to request the `draft` content for a `Collection`.

A bit of repetition in places, especially in `Collection`. I'd rather just make clear methods over doing something clever to work around `Option[List[Trail]]` and `List[Trail]`.

For this to truly work properly, it needs to be pointed to the draft content API; the user of the client needs to provide a `GuardianContentClient` that will do that.

@adamnfish @robertberry

### DIFF
--- a/fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
+++ b/fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
@@ -410,7 +410,7 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
       val faciaContent = FAPI.liveCollectionContentWithoutSnaps(collection)
 
       faciaContent.asFuture.futureValue.fold(
-        err => fail(s"expected 2 results, got $err", err.cause),
+        err => fail(s"expected 0 results, got $err", err.cause),
         contents => {
           contents.size should be(0)
         })
@@ -442,7 +442,7 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
       val faciaContentLive = FAPI.liveCollectionContentWithoutSnaps(collection, adjustSearchQuery = searchQuery => searchQuery.showTags("all"))
 
       faciaContent.asFuture.futureValue.fold(
-        err => fail(s"expected 2 results, got $err", err.cause),
+        err => fail(s"expected 4 results, got $err", err.cause),
         contents => {
           contents.size should be(4)
           contents.head.asInstanceOf[LatestSnap].latestContent.get.tags.exists(_.sectionId == Some("culture")) should be (true)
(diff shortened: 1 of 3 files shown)


fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
```diff
@@ -374,6 +374,92 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
...
+      faciaContent.asFuture.futureValue.fold(
+        err => fail(s"expected 2 results, got $err", err.cause),
+        contents => {
+          contents.size should be(2)
+          contents.head.asInstanceOf[LatestSnap].latestContent.get.tags.exists(_.sectionId == Some("culture")) should be (true)
+          contents.apply(1).asInstanceOf[LatestSnap].latestContent.get.tags.exists(_.sectionId == Some("technology")) should be (true)
+        })
+    }
+
+    "Should request a mix of both" in {
```
adamnfish: This looks like two different tests to me? You kick off two separate API calls and compare them separately? When you're using FreeSpec you can share test setup across multiple tests by just using an enclosing scope:

    "with my shared setup stuff" - {
      // shared setup here
      val shared = 5
      
      "test 1" in {
        // use `shared`
      }
      
      "test 2" in {
        // use `shared` again
      }
    }

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
```diff
@@ -374,6 +374,92 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
...
+      val normalTrailThree = Trail("internal-code/content/454695023", 0, None)
+      val dreamSnapOne = makeLatestTrailFor("snap/1281727", "uk/culture")
+      val dreamSnapTwo = makeLatestTrailFor("snap/2372382", "technology")
+      val collectionJson = makeCollectionJsonWithTreats(List(dreamSnapOne, normalTrail, dreamSnapTwo, normalTrailTwo), normalTrailThree)
+      val collection = Collection.fromCollectionJsonConfigAndContent("id", Some(collectionJson), collectionConfig)
+      val faciaContent = FAPI.draftCollectionContentWithSnaps(collection, adjustSnapItemQuery = itemQuery => itemQuery.showTags("all"))
+      val faciaContentLive = FAPI.liveCollectionContentWithoutSnaps(collection, adjustSearchQuery = searchQuery => searchQuery.showTags("all"))
+
+      faciaContent.asFuture.futureValue.fold(
+        err => fail(s"expected 2 results, got $err", err.cause),
```
adamnfish: This string (2) seems to misagree with the code below (4)? (see also below)

PR #97, opened 2015-04-29
ImageGallery for FaciaImage

janua:
This changes FaciaImage away from a `case class` into an `ADT`, which is what `ImageType` was.

On top of this, `ImageSlideshow` has been added to the type to represent `slideshows` from the tool.

In `facia-json`, there are now accessors for `imageSlideshowReplace` and `slideshow` which are keys in the `JSON` coming from the tool representing `slideshows`.

I have also removed the `toString` methods. If we want these as strings we should just make a method that says why it's turning into a `String`.

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
@@ -4,6 +4,11 @@ import play.api.libs.json._
 import org.joda.time.DateTime
 import play.api.libs.json.{JsValue, Json}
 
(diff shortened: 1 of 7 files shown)


adamnfish:
Much better, good change. I don't know much about the slideshow stuff but it seems legit?

janua:
@adamnfish Yep. `facia-tool` allows you to specify multiple images for a piece of content, which will turn it into a slideshow. The format of each object is the same as the `Replace` type, so I just used that.

robertberry:
:+1:

PR #98, opened 2015-05-07
Add isArticle to FaciaContentUtils

janua:
@adamnfish @robertberry 

This originally was to just add:
`def isArticle(fc: FaciaContent) = maybeContent(fc).exists(_.isArticle)`

But instead I also took the opportunity to refactor some of the code in `FaciaContentUtils`.


adamnfish:
Nice +/- ratio!

:+1:

PR #106, opened 2015-06-01
Fix draft requesting

janua:
We were not properly falling back to `live` when `draft` was a `None`.

This changes the signatures a bit to know that `draft` is a `None` and then fallback to `live` in the `None` case.

@robertberry

### DIFF
--- a/fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
+++ b/fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
@@ -432,6 +432,38 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
         })
     }
 
+    "should request what is in live when draft is None" - {
+      "for normal content" in {
+        val collectionJson = makeCollectionJson(normalTrail, normalTrailTwo)
+        val collection = Collection.fromCollectionJsonConfigAndContent("id", Some(collectionJson), collectionConfig)
+
+        val faciaContent = FAPI.draftCollectionContentWithoutSnaps(collection)
+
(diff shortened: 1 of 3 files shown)


robertberry:
@janua We should start gradually getting more client side people reviewing Scala pull requests ... I'm not here much longer :sandal:

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
```diff
@@ -432,11 +432,43 @@ class IntegrationTest extends FreeSpec with ShouldMatchers with ScalaFutures wit
...
+    "should request what is in live when draft is None" - {
+      "for normal content" in {
+        val collectionJson = makeCollectionJson(normalTrail, normalTrailTwo)
+        val collection = Collection.fromCollectionJsonConfigAndContent("id", Some(collectionJson), collectionConfig)
+
+        val faciaContent = FAPI.draftCollectionContentWithoutSnaps(collection)
+
+        faciaContent.asFuture.futureValue.fold(err => fail(s"expected 2 results, got $err", err.cause), contents => {
+          contents.size should be(2)
+          contents.head.asInstanceOf[CuratedContent].headline should be("PM returns from holiday after video shows US reporter beheaded by Briton")
```
robertberry: `asInstanceOf` you wot

robertberry:
:+1: otherwise

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
janua: I just want to check it. Now.

PR #110, opened 2015-06-05
Upgrade the contentAPI client to 6.3

janua:
I merged this, then reverted it when I realised how many problems will arise from it. I don't have the time to handle these problems now (They aren't problems in this repo, problems in `frontend`)

I will merge this when I am ready to receive it.

@DiegoVazquezNanini Is there anything in particular you want from `6.3`?


janua:
This no longer needs to be done because of #132.

PR #111, opened 2015-06-08
Carry frontPublicationDate through types

janua:
For breaking news, we would like to `frontPublicationDate` that exists on each `Trail` item in the `collection.json` files to be exposed in the higher level `fapi-client`.

We will put this into `lite.json`.

@stephanfowler @robertberry


stephanfowler:
:+1:

PR #112, opened 2015-06-08
Change Snap methods to fallback to LinkSnap 

janua:
#### Problem

There are legacy snaps hanging around our data without a `snapType`. After ~ Jan 2015, all snaps get a `snapType`, but some fronts (especially commercial fronts) have been static since long before this date.

#### Fix

This changes the `Snap` object methods `maybeFromTrail` and `maybeFromSupportingItem` to fallback to a `LinkSnap` when there is no `snapType` defined.

@stephanfowler @piuccio @robertberry


piuccio:
:+1:

robertberry:
:+1:

PR #113, opened 2015-06-09
Change Snap fallback to include new frontPublicationDate

janua:
These are now required since #111 got merged.


robertberry:
:+1:

PR #114, opened 2015-06-10
Revert snap fallback

janua:
This snap fallback has had unforeseen consequences around draft content.

We have decided to deal with the out-of-date data directly instead.


robertberry:
:+1:

PR #117, opened 2015-06-11
Fix CutOut resolving

janua:
When there was no `Content` in `getFaciaImage`, only `imageSlideShow` and `imageReplace` were getting called as fallbacks.

This adds in `imageCutout`. I have also simplified the code.

@robertberry


robertberry:
:+1:

PR #118, opened 2015-06-18
Fix SnapLatest Fields

janua:
This fixes fields in `LatestSnap` that were never being filled out properly like they were with the other types.

The fields changed are:
  - `href`
  - `byline`
  - `trailText`
  - `headline`

The only odd one is `href` in that it `FaciaContentUtils` it falls back to `snapUri`, whilst this `snapUri` can never end up in `href` on `LatestSnap` itself.

@robertberry


robertberry:
:+1:

PR #119, opened 2015-06-19
Add container label

NathanielBennett:
Create a description field for containers: Rendered thus is fronts tool. 

containter-field


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.50-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.51"
```
janua: You don't need to bump this

build.sbt
```diff
@@ -1,3 +1,4 @@
+
```
janua: You don't need this!

janua:
I don't understand what's going on between `CollectionConfigJson` and `CollectionJson`? Why are you adding a `description` to both?

janua:
Just a general note that my above comment has been resolved. I am going to merge this now.

PR #120, opened 2015-06-23
Switch ot internal page code

piuccio:
internal content code is now optional and will dispappear in the future

@janua


janua:
:+1:

PR #121, opened 2015-06-26
Playable live blog

piuccio:
Persist the new value created by guardian/frontend#9606

@stephanfowler @janua


stephanfowler:
:+1: ta!

PR #123, opened 2015-07-21
Use SpecialReport for tag "australia-news/series/healthcare-in-detention"

OliverJAsh:
I'm not sure if this should also be added to `common/app/views/support/CardStyleForFrontend.scala`?

We should put these behind a switch.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
@@ -23,7 +23,9 @@ object CardStyle {
     val href = trailMetaData.href.orElse(content.safeFields.get("href"))
     if (trailMetaData.snapType == Some("link") && href.exists(ExternalLinks.external)) {
       ExternalLink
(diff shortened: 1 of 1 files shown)


OliverJAsh:
Editors are currently hacking the system to special case these things:

http://preview.gutools.co.uk/australia-news/2015/jul/21/inevitable-over-asylum-seeker-health-targets-leaked-documents-show

image

The grey colour band and icon are done via an embedded interactive which scripts the page: http://interactive.guim.co.uk/next-gen/au/2015/jul/special-report-australia/shell/boot.js

/cc @theefer

theefer:
:expressionless:

OliverJAsh:
@theefer FYI, we are documenting hacks like these so we can accommodate for them if they become frequent enough.

theefer:
Good idea. Make sure to share the document with @na7hangood!

janua:
:+1:
No point merging until I am ready to release though. This will need a version bump on frontend after it gets released. Shall we just release it now?

OliverJAsh:
Yes, what is the workflow for releasing and version bumping?

PR #124, opened 2015-07-24
Remove editors picks from backfill results

janua:
TLDR; This forces `show-editors-picks` to be `false` on `backfill` queries.

### Problem

We are in a bit of a situation on some fronts that aren't massively curated and depend on backfills.

Step-by-step:
 - On frontend, when requesting backfills for a certain `path`, we ask for and **add** `editorsPicks` to the head of the `results` that `CAPI` returns to us.
 - We then generate a `lite.json` file from these results.
 - `CAPI` now reads this `lite.json` file and puts them into editors picks for the `path`.
 - On frontend, when requesting backfills for a certain `path`, we ask for and **add** `editorsPicks` to the head of the `results` that `CAPI` returns to us.

It's completely circular and confusing. We did sort this for some situations in the past, but this is what is left of it.

We have realised we definitely do not need `editorsPicks` anywhere near this library, as this library IS `editorsPicks`. Are you confused yet?

##### Why were we adding `editorsPicks` to the head of the list?

Adding `editorsPicks` to the head of the results like this worked fine and as expected up until we were the `editorPicks`. Before, `editorsPicks` came from `r2-admin`.

We always could see the circular dependency coming, and this is the last of it.

### Fix

Remove `editors-picks` from `facia-scala-client`. This will break the cycle.

@piuccio @stephanfowler @mchv

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 
+#### 0.54
+
+  - Changes queries to not add editorsPicks to the head of the list of backfill by default
+
+
 #### 0.53
+
   - Adds showLivePlayable content meta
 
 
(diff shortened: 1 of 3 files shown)


LATaylor-guardian:
Looks good from my side :+1: (And yes, very confusing!)

stephanfowler:
:+1:

LATaylor-guardian:
:+1:

PR #126, opened 2015-08-28
Metadata changes

janua:
This just adds two convenience methods: `ResolvedMetaData.toMap` and `CardStyle.fromContent`.

Also changes `ResolvedMetaData.fromContent` to a public method.


gklopper:
:+1:

PR #127, opened 2015-09-25
Group fronts

piuccio:
Group commercial fronts by 'something' that resemble desks, but not really.

Looks like this in fronts tool
screen shot 2015-09-25 at 16 58 58

@janua is this a breaking change?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+#### 0.57
+
+  - Adds Front.group meta
+
 #### 0.56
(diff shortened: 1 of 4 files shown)


janua:
If it's `Option` then no. But I would be really careful with it!

Update the `CHANGES`

piuccio:
changelog added, hopefully next version is `0.57` :smile:

janua:
:+1:

PR #128, opened 2015-09-28
Hide show more

piuccio:
Extra option on the collection to disable `show more` button even when there are extra trails.

TODO before merge
- [x] merge #127 
- [x] update `CHANGES.md`

TODO after merge
- [ ] update frontend to remove show more button
- [ ] [optional] update facia press to limit the number of trail
- [ ] update facia-tool to allow switching this value on/off
- [ ] rejoice

@janua

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
@@ -18,7 +18,8 @@ case class CollectionConfig(
     showDateHeader: Boolean,
     showLatestUpdate: Boolean,
     excludeFromRss: Boolean,
-    showTimestamps: Boolean)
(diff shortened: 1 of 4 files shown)


fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
```diff
@@ -54,5 +56,6 @@ object CollectionConfig {
       collectionJson.showDateHeader.getOrElse(false),
       collectionJson.showLatestUpdate.getOrElse(false),
       collectionJson.excludeFromRss.exists(identity),
-      collectionJson.showTimestamps.exists(identity))
+      collectionJson.showTimestamps.exists(identity),
+      collectionJson.hideShowMore.getOrElse(false))
```
janua: We should switch all these to `identity` really...

PR #131, opened 2015-10-21
Fix batch size

janua:
The maximum amount of `ids` that can be requested through a `search` query from the content API is 50.
This restricts the batch size to 50.

@piuccio

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/contentapi/IdsSearchQueries.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/contentapi/IdsSearchQueries.scala
@@ -7,10 +7,14 @@ object IdsSearchQueries {
   type Url = String
 
   val MaxUrlSize = com.gu.contentapi.client.Limits.UrlSize
+  //Maximum size that can be requested from CAPI is 50.
+  val MaxBatchSize = 50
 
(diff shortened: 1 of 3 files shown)


fapi-client/src/main/scala/com/gu/facia/api/contentapi/IdsSearchQueries.scala
```diff
@@ -7,10 +7,14 @@ object IdsSearchQueries {
   type Url = String
 
   val MaxUrlSize = com.gu.contentapi.client.Limits.UrlSize
+  //Maximum size that can be requested from CAPI is 50.
+  val MaxBatchSize = 50
```
piuccio: would be nice if `50` was exposed by `com.gu.contentapi.client.Limits.`

janua: True. But it's not. You should add it :smile:

piuccio:
:+1:

fapi-client/src/main/scala/com/gu/facia/api/contentapi/IdsSearchQueries.scala
LATaylor-guardian: @janua :laughing:

PR #132, opened 2015-11-03
Upgrade capi client to v7.3

davidfurey:
This also removes support for internalContentCode and so shouldn't be merged until tools have finished removing references based on internalContent from existing fronts.


LATaylor-guardian:
As for 'Convert CapiDate to Joda' - we have a helper for that. See: https://github.com/guardian/content-api-scala-client/blob/master/src/main/scala/com.gu.contentapi.client/utils/CapiModelEnrichment.scala

davidfurey:
Since no fronts are using internalContentCode anymore (http://count-content-code.herokuapp.com/), I think this can be merged?
@janua @piuccio @DiegoVazquezNanini

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -48,14 +49,14 @@ object ContentApi {
       // makes it possible to safely append additional tags
       case (k, v) if k == "tag" => (k, s"($v)")
       // ensure internalContentCode and internalPageCode are present on queries
```
LATaylor-guardian: This comment can be cleaned up now.

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -42,8 +42,7 @@ object Collection {
     // note that this does not currently deal with e.g. snaps
     def resolveTrail(trail: Trail): Option[FaciaContent] = {
       content.find { c =>
-        trail.id.endsWith("/" + c.safeFields.getOrElse("internalContentCode", "missing internal content code")) ||
-        trail.id.endsWith("/" + c.safeFields.getOrElse("internalPageCode", throw new RuntimeException("No internal code")))
+        trail.id.endsWith("/" + c.fields.flatMap(_.internalPageCode).getOrElse(throw new RuntimeException("No internal code")))
```
LATaylor-guardian: "no internal page code"

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -58,8 +57,7 @@ object Collection {
 
     def resolveSupportingContent(supportingItem: SupportingItem): Option[FaciaContent] = {
       content.find { c =>
-        supportingItem.id.endsWith("/" + c.safeFields.getOrElse("internalContentCode", "missing internal content code")) ||
-        supportingItem.id.endsWith("/" + c.safeFields.getOrElse("internalPageCode", throw new RuntimeException("No internal code")))
+        supportingItem.id.endsWith("/" + c.fields.flatMap(_.internalPageCode).getOrElse(throw new RuntimeException("No internal code")))
```
LATaylor-guardian: As above.

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
```diff
@@ -20,7 +20,7 @@ object CardStyle {
   val news = "news"
 
   def apply(content: Content, trailMetaData: MetaDataCommonFields): CardStyle = {
-    val href = trailMetaData.href.orElse(content.safeFields.get("href"))
```
piuccio: I don't understand the code much, but I've noticed this change in few places, aren't we missing something by getting rid of `content.safeFields.get("href")`?

davidfurey: I checked with @DiegoVazquezNanini and @janua, and we don't think that snap href will never be in content api.

fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
```diff
@@ -85,7 +85,7 @@ object ItemKicker {
 
   def seriesOrBlogKicker(item: Content) =
     item.tags.find({ tag =>
-      Set("series", "blog").contains(tag.`type`) && !TagsThatDoNotAutoKicker.contains(tag.id)
+      Set(TagType.Series: TagType, TagType.Blog).contains(tag.`type`) && !TagsThatDoNotAutoKicker.contains(tag.id)
```
janua: Just out of interest, why did you have to annotate the `TagType` here? Was scala not happy?

fapi-client/src/main/scala/com/gu/facia/api/utils/CardStyle.scala
janua: Yes, `href` doesn't seem to ever appear in the `fields` of a CAPI item. I think this has probably come from things of the past, where we wanted an items `href` to fallback to the content `href` (of some sort).
But in this case it doesn't seem to matter as we only use it to check if it's an external link and style is accordingly.

janua:
Great work! 
So glad to not be using arbitrary strings from CAPI anymore.
:+1:

LATaylor-guardian:
:+1: :)

fapi-client/src/main/scala/com/gu/facia/api/utils/ItemKicker.scala
davidfurey: Type mismatch, expected: TagType with Product, actual: TagType

janua: Weird typing with the types... it looks like it's mixing types. These are the ones from the thrift definitions I believe?
I wonder what would happen if you filled out the type of the set;
`Set[TagType](TagType.Series, TagType.Blog)`

davidfurey: That is effectively what annotating the type of the first element is doing, but better.  Thanks, I will change it

PR #135, opened 2015-11-18
Convert apiquery to backfill

piuccio:
Backfill is now an object with type and query. This allows future extensions to queries that are not backed by capi.

As a first step both `apiQuery` and `backfill` are available, the client tries to use `backfill` first and `apiQuery` if missing.

There's a change in a public method interface so I've created a new method with a different signature, users of the library have time to migrate.

@janua @DiegoVazquezNanini @davidfurey

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -5,7 +5,7 @@ import com.gu.contentapi.client.model.v1.Content
 import com.gu.facia.api.contentapi.ContentApi.{AdjustItemQuery, AdjustSearchQuery}
 import com.gu.facia.api.contentapi.{ContentApi, LatestSnapsRequest}
 import com.gu.facia.api.models._
-import com.gu.facia.api.utils.BackfillContent
+import com.gu.facia.api.utils.{BackfillResolver, BackfillContent}
 import com.gu.facia.client.ApiClient
 import com.gu.facia.client.models.TrailMetaData
 
@@ -182,7 +182,7 @@ object FAPI {
                          adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
                         (implicit capiClient: GuardianContentClient, faciaClient: ApiClient, ec: ExecutionContext): Response[List[FaciaContent]] = {
 
-    BackfillContent.resolveFromConfig(collection.collectionConfig)
-    .backfill(adjustSearchQuery, adjustItemQuery)
+    val resolver = BackfillContent.resolveFromConfig(collection.collectionConfig)
+    BackfillResolver.backfill(resolver, adjustSearchQuery, adjustItemQuery)
   }
 }
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -182,7 +182,7 @@ object FAPI {
                          adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
                         (implicit capiClient: GuardianContentClient, faciaClient: ApiClient, ec: ExecutionContext): Response[List[FaciaContent]] = {
 
-    val resolver = BackfillResolver.resolveFromConfig(collection.collectionConfig)
-    BackfillResolver.backfill(resolver, adjustSearchQuery, adjustItemQuery)
+    val backfillRequest = BackfillResolver.resolveFromConfig(collection.collectionConfig)
+    BackfillResolver.backfill(backfillRequest, adjustSearchQuery, adjustItemQuery)
   }
 }
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -5,7 +5,7 @@ import com.gu.contentapi.client.model.v1.Content
 import com.gu.facia.api.contentapi.ContentApi.{AdjustItemQuery, AdjustSearchQuery}
 import com.gu.facia.api.contentapi.{ContentApi, LatestSnapsRequest}
 import com.gu.facia.api.models._
-import com.gu.facia.api.utils.{BackfillResolver, BackfillContent}
+import com.gu.facia.api.utils.BackfillResolver
 import com.gu.facia.client.ApiClient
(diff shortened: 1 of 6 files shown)


piuccio:
I've update this PR, it now has a `BackfillResover` that implements two types of backfill query

* `capi`: the traditional capi search string
* `collection`: inherit the **curated content** of a parent collection

I don't think the child collection should inherit the parent's backfill, at least until we solve loops and deduping.

@janua I've added unit tests!

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -0,0 +1,63 @@
...
+import com.gu.facia.api.models.{FaciaContent, CollectionConfig, CuratedContent}
+import com.gu.facia.client.ApiClient
+import com.gu.facia.client.models.{CollectionJson, Backfill, TrailMetaData}
+
+import scala.concurrent.ExecutionContext
+
+object BackfillContent {
+  def resolveFromConfig(collectionConfig: CollectionConfig): BackfillResolver = {
+    collectionConfig.backfill match {
+      case Some(backfill: Backfill) => backfill.`type` match {
```
janua: Switch to nested case matching here.

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -0,0 +1,63 @@
...
+  def backfill(adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
+              (implicit capiClient: GuardianContentClient, faciaClient: ApiClient, ec: ExecutionContext): Response[List[FaciaContent]] = {
+    for {
+      collection <- FAPI.getCollection(parentCollectionId)
+      curatedCollection <- FAPI.liveCollectionContentWithSnaps(collection, adjustSearchQuery, adjustItemQuery)
+    } yield curatedCollection
+  }
+}
+
+case class EmptyBackfill() extends BackfillResolver {
```
janua: This can be an `object`.

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import scala.concurrent.ExecutionContext
+
+object BackfillContent {
+  def resolveFromConfig(collectionConfig: CollectionConfig): BackfillResolver = {
+    collectionConfig.backfill match {
+      case Some(backfill: Backfill) => backfill.`type` match {
+        case "capi" => CapiBackfill(backfill.query, collectionConfig)
+        case "collection" => CollectionBackfill(backfill.query)
+        case _ => throw new Error(s"Invalid backfill type ${backfill.`type`}")
```
janua: Are you sure you want an `Exception` here? Why not `EmptyBackfill`?
If so, write a quick unit test for it to show that's what is expected.

piuccio:
thanks @janua I've fixed your comments.

I prefer to throw an exception when the backfill is invalid because it's a configuration error, it's something that shouldn't happen. Returning an empty list will silently ignore an invalid state. I've added a test for that

piuccio:
@janua I've updated the PR with a better (???) backfill resolver

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.gu.facia.api.{FAPI, Response}
+import com.gu.facia.client.ApiClient
+import com.gu.facia.client.models.{Backfill, TrailMetaData}
+
+import scala.concurrent.ExecutionContext
+
+case class InvalidBackfillConfiguration(msg: String) extends Error(msg)
+
+object BackfillContent {
+  def resolveFromConfig(collectionConfig: CollectionConfig): BackfillResolver = {
```
janua: This could now go into `BackfillResolver`!

Is the name `BackfillContent` better than `BackfillResolver`?

piuccio: they are now in the same companion object

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -0,0 +1,53 @@
...
+import com.gu.facia.api.contentapi.ContentApi
+import com.gu.facia.api.contentapi.ContentApi._
+import com.gu.facia.api.models.{CollectionConfig, CuratedContent, FaciaContent}
+import com.gu.facia.api.{FAPI, Response}
+import com.gu.facia.client.ApiClient
+import com.gu.facia.client.models.{Backfill, TrailMetaData}
+
+import scala.concurrent.ExecutionContext
+
+case class InvalidBackfillConfiguration(msg: String) extends Error(msg)
```
janua: Don't extend `Error`. Instead extend `Exception`.

`Error` is used for critical errors such as `ThreadDeath` and `VirtualMachineError`.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -170,4 +172,17 @@ object FAPI {
...
+  /**
+    * Fetches content for the configured backfill query. The query can be manipulated for different
+    * requirements by providing adjustment functions. The results then have their facia metadata
+    * resolved using the collection information.
+    */
+  def backfillFromConfig(collection: Collection,
+                         adjustSearchQuery: AdjustSearchQuery = identity, adjustItemQuery: AdjustItemQuery = identity)
+                        (implicit capiClient: GuardianContentClient, faciaClient: ApiClient, ec: ExecutionContext): Response[List[FaciaContent]] = {
+
+    val resolver = BackfillResolver.resolveFromConfig(collection.collectionConfig)
```
janua: `backfillRequest`? `resolver` means nothing to me for this `val`!

janua:
Looks good :clap: :+1:

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
piuccio: fixed :smile:

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
piuccio: fixed :smile:

PR #136, opened 2015-11-30
update dependencies

piuccio:
and instead of depending on the whole `aws-java-sdk` only include `s3`

@janua


janua:
:+1:
Cool, it used to be all or nothing.

rich-nguyen:
Amazing! I'm glad you've done this, because it helps cut down our frontend artifact size

PR #142, opened 2016-01-27
update dependencies

piuccio:
I've created a new front called `fapi-integrationtest-...` and updated tests because the collection I was using for testing got deleted.

@janua


janua:
:+1:

PR #143, opened 2016-02-15
add a tag attribute to CollectionJson

Reettaphant:
@janua @piuccio

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -18,17 +18,19 @@ case object Special extends Metadata
 
 case object Breaking extends Metadata
 
+
 object Metadata {
 
-  val tags = List("Canonical", "Breaking", "Special")
+  val tags: Map[String, Metadata] = Map(("Canonical", Canonical), ("Special", Special), ("Breaking", Breaking))
 
   implicit object MetadataFormat extends Format[Metadata] {
     def reads(json: JsValue) = {
       (json \ "type").transform[JsString](Reads.JsStringReads) match {
-        case JsSuccess(JsString("Canonical"), _) => JsSuccess(Canonical)
-        case JsSuccess(JsString("Special"), _) => JsSuccess(Special)
-        case JsSuccess(JsString("Breaking"), _) => JsSuccess(Breaking)
-        case _ => JsError("Could not convert CollectionTag")
+        case JsSuccess(JsString(string), _) => tags.get(string) match {
+          case Some(result) => JsSuccess(result)
+          case _ => JsError("Could not convert CollectionTag: string is of unknown type")
+        }
+        case _ => JsError("Could not convert CollectionTag: type is not a string")
       }
     }
 
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -1,6 +1,6 @@
 package com.gu.facia.client.models
 
-import play.api.libs.json.Json
+import play.api.libs.json._
 
 object Backfill {
   implicit val jsonFormat = Json.format[Backfill]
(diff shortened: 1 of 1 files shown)


facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
```diff
@@ -112,6 +112,30 @@ case class Trail(
   lazy val safeMeta = meta.getOrElse(TrailMetaData.empty)
 }
 
+sealed trait CollectionTag {}
+
+case object Tag1 extends CollectionTag {}
+
+case object Tag2 extends CollectionTag {}
```
janua: You don't need any of these brackets

janua:
Should they definitely be called `Tag1` and `Tag2`? It seems a bit vague to me!

Reettaphant:
I don't actually have any real tag names yet. This should probably wait until there is at least one.

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -10,6 +10,33 @@ case class Backfill(
...
 
+sealed trait MetaData
+
+case object Tag1 extends MetaData
+
+case object Tag2 extends MetaData
+
+object MetaData {
+
+  implicit val tags = List("Tag1", "Tag2")
```
piuccio: why implicit? and shouldn't it be `List(Tag1, Tag2)` ?

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -10,6 +10,33 @@ case class Backfill(
...
+
+  implicit object MetaDataFormat extends Format[MetaData] {
+    def reads(json: JsValue) = {
+      (json \ "type").transform[JsString](Reads.JsStringReads) match {
```
piuccio: can't we just serialize to the string `Tag1` instead of `{ type: "Tag1" }`?

janua: It's better to stick to a `type` field in this case, as it is an object and not a string.

You could do this, but then how do you tell the difference between a string and an object, and how do you add more fields to the object (if you needed it).

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
janua: I don't think this is used; I can't see what would use this implicit.

piuccio:
The final list of tags is `Canonical`, `Special` and `Breaking`

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -19,6 +50,7 @@ object CollectionConfigJson {
     displayName: Option[String] = None,
     apiQuery: Option[String] = None,
     backfill: Option[Backfill] = None,
+    tags: Option[List[MetaData]] = None,
```
piuccio: shouldn't this be called `metadata`?

piuccio:
@janua this looks good to me, what do you think?

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -10,6 +10,37 @@ case class Backfill(
...
+
+case object Canonical extends Metadata
+
+case object Special extends Metadata
+
+case object Breaking extends Metadata
+
+object Metadata {
+
+  val tags = List("Canonical", "Breaking", "Special")
```
janua: What's the point of `tags`?

Reettaphant: Facia tool needs a list of all available tags.

janua: Maybe it would be more useful as a mapping to the type; `Map[String, Metadata]`. You could then use it in the `reads` for example.

Reettaphant:
@janua I've made your suggested changes now.

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -10,6 +10,39 @@ case class Backfill(
...
+  val tags: Map[String, Metadata] = Map(("Canonical", Canonical), ("Special", Special), ("Breaking", Breaking))
+
+  implicit object MetadataFormat extends Format[Metadata] {
+    def reads(json: JsValue) = {
+      (json \ "type").transform[JsString](Reads.JsStringReads) match {
+        case JsSuccess(JsString(string), _) => tags.get(string) match {
+          case Some(result) => JsSuccess(result)
+          case _ => JsError("Could not convert CollectionTag")
+        }
+        case _ => JsError("Could not convert CollectionTag")
```
janua: Better error messages for these?
`type` is not a string and `string` is of unknown type.

janua:
:+1:

PR #144, opened 2016-02-18
Remove apiQuery

piuccio:
This is a breaking change, remove a deprecated method and update all dependencies. MAPI and frontend are both using Play 2.4 so there's no need to be outdated

@janua this deserves a `1.0.0`


janua:
Why don't you split this into two?

piuccio:
done in #145

piuccio:
updated

screen shot 2016-02-24 at 12 40 32

piuccio:
renamed the PR :grin:  it's basically a bump :tongue:

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -153,27 +152,6 @@ object FAPI {
     yield Collection.draftContent(collection, setOfContent, snapContent)}
 
   /**
-   * Fetches content for the given backfill query. The query can be manipulated for different
-   * requirements by providing adjustment functions. The results then have their facia metadata
-   * resolved using the collection information.
-   */
-  @deprecated
-  def backfill(backfillQuery: String, collection: Collection,
```
janua: Do you need to remove this in this PR?

Seems like you could just do the rename without removing any deprecated functions, unless this depends on `apiQuery` and won't work without it?

piuccio: `backfillQuery` should not be exposed to users of the library. This method is made obsolete by  `backfillFromConfig`

piuccio: > Do you need to remove this in this PR?
Yes, we deprecated `backfill` in 0.68, frontend and apps upgraded fapi, no one is using that method anymore, so it should be removed because there's another one that hides the implementation details of backfill.

janua:
:+1:

PR #146, opened 2016-02-24
downgrade test dependency

piuccio:
because it conflicts with specs2

I'm sure I run the tests when bumping dependencies, but for some reason now tests fail on master.
Tracked down to some incompatible transitive dependencies...

This should work (or at least it just worked here).

screen shot 2016-02-24 at 11 07 38


janua:
:+1:

PR #150, opened 2016-03-14
Remove backfill errors 

Reettaphant:
This pull request fixes two things: 

- If a collection has a backfill of type collection where the collectionId refers to a parent collection that does not exist, don't return an error but just return an empty list for the backfill instead

- If a backfill has a backfill type that is not recognised, don't throw an error but return an empty backfill.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
@@ -22,7 +22,7 @@ object BackfillResolver {
     collectionConfig.backfill match {
       case Some(Backfill("capi", query: String)) => CapiBackfill(query, collectionConfig)
       case Some(Backfill("collection", query: String)) => CollectionBackfill(query)
-      case Some(Backfill(backFillType, _)) => throw new InvalidBackfillConfiguration(s"Invalid backfill type $backFillType")
+      case Some(Backfill(backFillType, _)) => EmptyBackfill
       case None => EmptyBackfill
     }
   }
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
@@ -40,10 +40,16 @@ object BackfillResolver {
         } yield {
           backfillContent.map(CuratedContent.fromTrailAndContent(_, TrailMetaData.empty, None, collectionConfig))
         }
-      case CollectionBackfill(parentCollectionId) =>
-        for {
+      case CollectionBackfill(parentCollectionId) => {
+       val collectionBackfillResult =
+         for {
           collection <- FAPI.getCollection(parentCollectionId)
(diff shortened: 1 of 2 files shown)


piuccio:
Nice, that's how I should have implemented in the first place.

The parent collection might disappear at any point, this shouldn't prevent other fronts from pressing.

Returning empty list when backfill is not recognized means facia tool can create new types without breaking frontend

@janua

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -22,8 +22,7 @@ object BackfillResolver {
     collectionConfig.backfill match {
       case Some(Backfill("capi", query: String)) => CapiBackfill(query, collectionConfig)
       case Some(Backfill("collection", query: String)) => CollectionBackfill(query)
-      case Some(Backfill(backFillType, _)) => throw new InvalidBackfillConfiguration(s"Invalid backfill type $backFillType")
-      case None => EmptyBackfill
+      case _ => EmptyBackfill
```
janua: Could we avoid using `_` in place of just using `EmptyBackfill` twice?

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -40,10 +39,11 @@ object BackfillResolver {
         } yield {
           backfillContent.map(CuratedContent.fromTrailAndContent(_, TrailMetaData.empty, None, collectionConfig))
         }
-      case CollectionBackfill(parentCollectionId) =>
+      case CollectionBackfill(parentCollectionId) => {
         for {
           collection <- FAPI.getCollection(parentCollectionId)
           curatedCollection <- FAPI.liveCollectionContentWithSnaps(collection, adjustSearchQuery, adjustItemQuery)
         } yield curatedCollection
+      } recover { err => List() }
```
janua: This feels a bit hidden and easy to miss! I think you should put the future into it's own `val` and then do the recover in that. And in there add logging, otherwise we will never know how often this happens.

Reettaphant: @janua Ok, I'll do this. Can you give me an example of where logging is done in facia tool?

fapi-client/src/main/scala/com/gu/facia/api/utils/Backfill.scala
```diff
@@ -40,10 +40,18 @@ object BackfillResolver {
...
+         for {
           collection <- FAPI.getCollection(parentCollectionId)
           curatedCollection <- FAPI.liveCollectionContentWithSnaps(collection, adjustSearchQuery, adjustItemQuery)
         } yield curatedCollection
+        
+        collectionBackfillResult recover {
+          err => List()
+        }
+
+        collectionBackfillResult
```
janua: This ignores the `recover`! The `recover` needs to be the last statement.

janua:
:+1:

PR #151, opened 2016-03-23
create a new imageSource property for imageReplace

Reettaphant:
Add a new imageSource property to metadata that contains imageSource and imageSourceOrigin. This will be used for imageReplace logic in the future. @piuccio


facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
```diff
@@ -9,6 +9,11 @@ object SlideshowAsset {
   implicit val slideshowAssetFormat = Json.format[SlideshowAsset]
 }
 
+case class ImageSourceAsset(imageSrc: String, imageSrcOrigin: String)
```
piuccio: in the model these are called `src` and `origin`

fapi-client/src/main/scala/com/gu/facia/api/models/curatedcontent.scala
```diff
@@ -8,6 +8,7 @@ sealed trait FaciaImage
 case class Cutout(imageSrc: String, imageSrcWidth: Option[String], imageSrcHeight: Option[String]) extends FaciaImage
 case class Replace(imageSrc: String, imageSrcWidth: String, imageSrcHeight: String) extends FaciaImage
 case class ImageSlideshow(assets: List[Replace]) extends FaciaImage
+case class ImageReplace(imageSrc: String, imageSrcOrigin: String) extends FaciaImage
```
piuccio: the curated content doesn't have to know about `imageSrcOrigin`

PR #157, opened 2016-04-07
Publish a separate artifact for play-json 2.5.x

cb372:
It seems that play-json 2.4.6 doesn't play nicely (excuse the pun) when you use the lib in a Play 2.5.1 app.


cb372:
Merging, as nobody seems to disagree with this approach.

PR #160, opened 2016-04-13
Added note about ImageReplace

rich-nguyen:
Update the changes.md doc to include @Reettaphant ImageReplace class. I didn't realise that had been added when I published...


piuccio:
👍

PR #162, opened 2016-05-09
Enable testing using AWS session token

kelvin-chappell:
This adds a named AWS profile to the credentials provider chain in the integration tests and demo worksheet.

### DIFF
--- a/fapi-client/src/test/scala/demo.sc
+++ b/fapi-client/src/test/scala/demo.sc
@@ -9,20 +9,23 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 
 // demo config
-implicit val capiClient = {
-  val apiKey: String = scala.sys.env.getOrElse("CONTENT_API_KEY", "")
-  new GuardianContentClient(apiKey)
+val apiKey = "CONTENT_API_KEY"
+val awsProfileName = Some("AWS_PROFILE_NAME")
+
+implicit val capiClient =  new GuardianContentClient(apiKey)
+implicit val apiClient: ApiClient = {
+  val amazonS3Client = {
+    val credentialsProvider = new AWSCredentialsProviderChain(
+      new EnvironmentVariableCredentialsProvider(),
+      new SystemPropertiesCredentialsProvider(),
+      awsProfileName map {
+        new ProfileCredentialsProvider(_)
+      } getOrElse new ProfileCredentialsProvider()
+    )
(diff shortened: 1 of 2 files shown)


fapi-client/src/test/scala/demo.sc
```diff
@@ -1,24 +1,38 @@
...
+val apiKey = "CONTENT_API_KEY"
+val awsProfileName = Some("AWS_PROFILE_NAME")
+
+implicit val capiClient =  new GuardianContentClient(apiKey)
+implicit val apiClient: ApiClient = {
+  val amazonS3Client = {
+    val credentialsProvider = new AWSCredentialsProviderChain(
+      new EnvironmentVariableCredentialsProvider(),
+      new SystemPropertiesCredentialsProvider(),
+      awsProfileName map {
```
janua: Why the `map`? You could just change `awsProfileName` to a `String` and use `ProfileCredentialsProvider` directly.

janua:
What's stopping you from using a named `profile` rather than using an environment variable to find out the named `profile`?

fapi-client/src/test/scala/demo.sc
kelvin-chappell: It was to allow the default profile to be used if you don't give a profile name.

janua: But in this demo it's always going to be `AWS_PROFILE_NAME` ?

kelvin-chappell:
I think there's more than one profile that has access, and could be using the default profile.  Also wasn't sure what teamcity was using, if anything.

fapi-client/src/test/scala/demo.sc
adamnfish: I think it's probably better to be explicit, especially given that our dev workflow with Janus means there'll always be a profile name. As cross-team (/account) collaboration gets more common this'll become more pronounced and avoiding default profiles in our workflows will make it easier for devs to pick up services outside their day-to-day work.

kelvin-chappell: That means *fill value in here*
Maybe I should call it `FILL_IN_VALUE_HERE`
I see what you mean about the map now! I'll get rid of that.

janua:
I thought the profile we would want to use by default would be `cmsFronts` ?

kelvin-chappell:
I've now updated the profile to `cmsFronts` for both.

janua:
Looks good :+1:

PR #163, opened 2016-05-09
Add branded tag to collections model

kelvin-chappell:
This is to label containers that should have a logo or other commercial branding.

/cc @piuccio


piuccio:
looks good

PR #164, opened 2016-05-16
Add metadata to collection config

kelvin-chappell:


fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
```diff
@@ -1,12 +1,13 @@
...
 
-import com.gu.facia.client.models.{Backfill, CollectionConfigJson}
+import com.gu.facia.client.models.{Backfill, CollectionConfigJson, Metadata}
 
 case class Groups(groups: List[String])
 
 case class CollectionConfig(
     displayName: Option[String],
     backfill: Option[Backfill],
+    metadata: Option[List[Metadata]],
```
janua: Why is this `List` of `Metadata`? I thought the original idea was it to be categorised as one of them.

kelvin-chappell: A collection can have many of them.  They appear as `tags` in the fronts tool.
Eg:
image

janua: Ah great! I haven't used them. In the past for canonical it could only be one.

janua:
@kelvin-chappell Kind of surprised that this hasn't already been exposed? What do these tags currently do when use them? They can't be doing anything on `frontend` for example.

Looks good :+1:

kelvin-chappell:
@janua I don't think these tags do anything yet.  This is the cutting edge!

PR #166, opened 2016-05-17
Add UnknownMetadata

janua:
This adds `UnknownMetadata` to the `Metadata` ADT.

This means when we add new tags to `Metadata`, we don't break old clients. Clients will have to correctly handle `UnknownMetadata` after this change.

This **also** adds logging via `logback` and `scala-logging` to log out the case of falling back to `UnknownMetadata`; we will also use logging in other parts of the library also.

@piuccio @kelvin-chappell


piuccio:
nice 👍

kelvin-chappell:
Good idea 👍

PR #167, opened 2016-06-13
Add width and height to image replace

piuccio:
Some time ago we added a new image replace object, including only src and origin. Turns out that iphone requires width and height to compute the aspect ratio

This change adds width and height to facia-json (the low level layer) and remove ImageReplace from the upper level.
Case class Replace is capable of representing all the information needed by applications.

@janua  @Reettaphant


janua:
Looks good! :+1: 
Don't forget this will break any current `ImageSourceAsset` that exist without a `width` and `height`. (It will at least safely come out as `None`, instead of actually breaking any applications).

piuccio:
`ImageSourceAsset` is not yet used by anyone (except for a test on CODE), should be safe to change

PR #169, opened 2016-07-14
construct the short code more robustly

johnduffell:
we need the short code for discussion comment requests.  CAPI doesn't supply that (at present) so we try to work it out by stripping http://gu.com from the front of the short url.  Now that they're https it doesn't work, so comment counts are not working in prod due to the data-discussion-id still having the https part.

Perhaps in reality discussion should be happy with long urls? or CAPI should supply the discussion key? @guardian/discussion @guardian/content-platforms 

Anyway for now I've just made the regex more generous.  But fabio said he's noticed it hard coded in a few places.

Once I've merged this one I'll ship a new version of fapi-client and ship a frontend PR to use it.

@piuccio @janua

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
@@ -136,7 +136,7 @@ object FaciaContentUtils {
   )
   def maybeShortUrl(fc: FaciaContent) = fieldsGet(fc)(_.flatMap(_.shortUrl))
   def shortUrl(fc: FaciaContent): String = maybeShortUrl(fc).getOrElse("")
-  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replace("http://gu.com", ""))
+  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replaceFirst("[a-z]+://gu.com", ""))
   def discussionId(fc: FaciaContent) = shortUrlPath(fc)
 
   def isBoosted(fc: FaciaContent): Boolean = fold(fc)(
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
@@ -136,7 +136,7 @@ object FaciaContentUtils {
   )
   def maybeShortUrl(fc: FaciaContent) = fieldsGet(fc)(_.flatMap(_.shortUrl))
   def shortUrl(fc: FaciaContent): String = maybeShortUrl(fc).getOrElse("")
-  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replaceFirst("[a-z]+://gu.com", ""))
+  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replaceFirst("^[a-zA-Z]+://gu.com", ""))
   def discussionId(fc: FaciaContent) = shortUrlPath(fc)
 
   def isBoosted(fc: FaciaContent): Boolean = fold(fc)(


mchv:
@johnduffell We have discussed about this and are a bit reluctant to add a new field for legacy reasons, but maybe it should just be an internal field.

johnduffell:
yeah at the moment effectively there's a foreign key relationship between capi and dapi which requires all clients to do text processing on another field to find out the key.  I'd be happy enough if it was called discussion-key, probably even more so, as long as it's opaque to clients.

gidsg:
Can we merge this as-is and then replace it with a new field if/when that appears?

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
```diff
@@ -136,7 +136,7 @@ object FaciaContentUtils {
   )
   def maybeShortUrl(fc: FaciaContent) = fieldsGet(fc)(_.flatMap(_.shortUrl))
   def shortUrl(fc: FaciaContent): String = maybeShortUrl(fc).getOrElse("")
-  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replace("http://gu.com", ""))
+  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replaceFirst("^[a-zA-Z]+://gu.com", ""))
```
adamnfish: Would it be more robust to use a URL parser on this? You're after the "path". That way it'll work whatever changes around it (maybe at some point they'll start adding querystrings here, for example).

```scala
import java.net.URL

def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(new URL(_).getPath)
```

johnduffell: I think the problem is there's no defined way to get a discussion key from a capi item.  That might work, but I don't know why they might add a query string, and whether that query string might be in or out of the discussion key.

adamnfish: Regular Expressions

mchv: I am adding the key here https://github.com/guardian/content-api-models/pull/45

johnduffell: oh come on adam, regexes are like hammers, you can use them on anything! 🔨

dominickendrick: would `replaceFirst("^https?://gu.com", "")` be clearer ?

janua: What @dominickendrick said. And what Adam said more.
Although I think it doesn't matter now that the `discussionKey` is going to be coming through?

johnduffell: @dominickendrick it would, but I noticed in other places in frontend we use what I wrote, so I just copied and pasted that, but yeah, what everyone else said

nicl:
@johnduffell @mchv as @gidsg says, can we merge to get this working and then work out a better solution?

johnduffell:
anyone happy to give me a +1?  if people really want I can use a different regex, or a url parser, but it seems unnecessary.  I'll be begging for +1s in person tomorrow morning otherwise!

mchv:
👍

mchv:
By the way content api scala client `9.12` as added a new optional field `internalShortId` that contains the id used by discussion as a key (e.g `/p/4pvzj`).

We have not yet fully reindex so old articles does not have it, but any (re)published article will have  it.

johnduffell:
thanks @mchv let me know when the index is complete and I can do another release

PR #171, opened 2016-08-12
document nauru files tag

johnduffell:
Now that it is public, we can document it.

No idea what the one above it is though :(

We should probably have something like the Badge/SpecialBadge system where we only hash the ones that aren't public... or much better still grab the list from S3 in an agent.  If we do switch, we won't actually know what to put in for the mystery hash...

@TBonnin


janua:
@johnduffell Yeah, I think we should refactor this part of the code into something more typed. Essentially having non hashed AND hashed versions that can be checked against. And it shouldn't be inline like it is here on `CardStyle` anymore.

:+1:

johnduffell:
@janua I meant to go and grab the link to Badge/SpecialBadge but I forgot: https://github.com/guardian/frontend/blob/master/common/app/model/Badges.scala#L11

PR #173, opened 2016-10-11
Bump CAPI client and add better logging in ContentApi

janua:
This bumps the CAPI client to `10.5`.

This also adds better logging into the `ContentApi` object; warning when a response may be attempted to be construed as an `ItemResponse` when it is actually a  `SearchResponse`.

I've also added a few lose types.

@gtrufitt @LATaylor-guardian

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+#### 2.0.4
+
+  - Upgrade CAPI client to 10.5
+  - Add better logging in ContentApi object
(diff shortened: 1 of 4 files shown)


johnduffell:
> #173 opened on Oct 11 by janua  

>This branch has conflicts that must be resolved

😱

buck06191:
CAPI client is way beyond this now. Closing as old

PR #175, opened 2016-11-28
Add functions to inject thumbnail urls into a Facia collection.

philmcmahon:
Currently when replacing an image using the fronts tool, the master version of the image is used as the thumbnail. This is problematic for people in Australia who end up loading a massive file to display a tiny image.

The grid API doesn't provide permanent thumbnail urls, so it doesn't make sense to store them in the fronts datastore, so here I've taken the approach of injecting generated secure URLs into the api response (PR on facia tool to follow).


janua:
@philmcmahon Why does this code have to live in `facia-scala-client` if it is unused by the library?

PR #177, opened 2016-12-05
set showMainVideo to for Video Atoms

gidsg:
If video page contains atom main media set `showMainVideo` to false, else use previous behaviour.

*Perhaps not the most elegant way but we found it achieves the intended business logic*

CC @janua @Reettaphant

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
@@ -1,8 +1,8 @@
 package com.gu.facia.api.utils
 
-import com.gu.contentapi.client.model.v1.Content
 import com.gu.facia.client.models.MetaDataCommonFields
 import play.api.libs.json.{Format, Json}
+import com.gu.contentapi.client.model.v1.{Content, Element, ElementType}
 
 
 object ResolvedMetaData {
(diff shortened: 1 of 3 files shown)


fapi-client/src/test/scala/com/gu/facia/api/utils/ResolvedMetaDataTest.scala
```diff
@@ -1,67 +1,64 @@
...
       "showQuotedHeadline" -> JsBoolean(true),
       "imageCutoutReplace" -> JsBoolean(true)))
-
   val trailMetaDataWithFieldsSetFalse = TrailMetaData(
     Map("showByline" -> JsBoolean(false),
       "showMainVideo" -> JsBoolean(false),
       "showQuotedHeadline" -> JsBoolean(false),
       "imageCutoutReplace" -> JsBoolean(false)))
 
+  def tagWithId(id: String) = Tag(
```
janua: Why has this moved? Seems unnecessary!

gidsg: I think this was some IntelliJ reformat or somesuch, I can revert.

fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
```diff
@@ -21,6 +21,18 @@ object ResolvedMetaData {
   def isVideoForContent(content: Content): Boolean =
     content.tags.exists(_.id == Video)
 
+  def isVideoAtom(content: Content): Boolean = {
+    val maybeAtom = for {
+      videoContentType <- content.tags.find(_.id == Video)
+      blocks <- content.blocks
+      main <- blocks.main
+      atomElement <- main.elements.find(e => e.`type` == ElementType.Contentatom)
+    } yield atomElement
```
janua: This `for` comprehension is tricky to parse if you aren't reading it carefully.

Could you either add the type of `Option[List[Option[Element]]` or change it around to make it easier to read? I think it would be clearer what you are trying to do by using `exists` and then not having to use the `isDefined` on the big nested type.

janua:
Great! 👍 
Let's get this out and the get frontend, tools and mapi onto the new version before their usage.

janua:
[Helping Youtube implementation on NextGen](https://trello.com/c/SEVhjRxu/222-helping-youtube-implementation-on-nextgen)

gidsg:
@janua please can you merge this and deploy to Sonatype, then I will update its clients.

janua:
@gidsg `2.0.6` released!

PR #178, opened 2016-12-20
default show main video to true for video atom pages

gidsg:
defaults show main video to true for video atom pages. This is in preparation for adding playable support to frontend. 

This is restricted to YouTube atoms, so it will still be false for other types of media atoms.

CC @Reettaphant @janua

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
@@ -2,7 +2,11 @@ package com.gu.facia.api.utils
 
 import com.gu.facia.client.models.MetaDataCommonFields
 import play.api.libs.json.{Format, Json}
-import com.gu.contentapi.client.model.v1.{Content, Element, ElementType}
+import com.gu.contentapi.client.model.v1.{Content, ElementType}
+import com.gu.contentatom.thrift.AtomData
+import com.gu.contentatom.thrift.atom.media.Platform
+
+import scala.util.Try
 
 
 object ResolvedMetaData {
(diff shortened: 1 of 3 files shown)


fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
```diff
@@ -18,17 +22,21 @@ object ResolvedMetaData {
...
-      videoContentType <- content.tags.find(_.id == Video)
-      blocks <- content.blocks
-      main <- blocks.main
-    } yield main.elements.exists(e => e.`type` == ElementType.Contentatom)
 
-    atomExists.getOrElse(false)
+  def isYouTubeMediaAtomForContent(content: Content): Boolean = {
+    val youTubeAtomExists: Option[Boolean] = for {
+      videoContentType <- content.tags.find(_.id == Video)
+      atoms <- content.atoms
```
janua: `atom` ?

fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
```diff
@@ -18,17 +22,21 @@ object ResolvedMetaData {
   def isCommentForContent(content: Content): Boolean =
     content.tags.exists(_.id == Comment)
 
-  def isVideoForContent(content: Content): Boolean =
-    content.tags.exists(_.id == Video)
+  def isVideoElementForContent(content: Content): Boolean =
+    content.tags.exists(_.id == Video) && content.elements.exists(_.exists(_.`type` == ElementType.Video))
```
janua: Why are we now adding this extra restriction?
It's up to the client to request the data, but I don't think we should set defaults based on whether the client has requested it or not. Rather, I think we should always set the default and it's up to the client to render it appropriately.
Although maybe we could change how it works and do it like this now instead?

gidsg:
after extensive chatting IRL with @janua we have decided it would be better to revert to the previous lazy true assumption that all `type/video` content defaults to true. 

The downside of this is that some content such as new atom media types may default to true without the ability to turn it off in facia-tool. The upside is simpler logic in facia-scala-client.

PR #181, opened 2017-01-09
Word count

davidfurey:
Adds method to get wordcount for content.  This is so we can test adding an approximate reading time indication on fronts.


janua:
:+1:

PR #186, opened 2017-04-20
Label paid content accordingly

marialivia16:
Paid content in facia tool appears as 'news' and it needs to be labelled to indicate that this is 'paid' content instead.

This has been tested with the `facia-tool`.


marialivia16:
[Paid content in facia tool appears as 'news' could it be labelled to indicate that this is 'paid' content instead?](https://trello.com/c/C9ECZJdV/53-paid-content-in-facia-tool-appears-as-news-could-it-be-labelled-to-indicate-that-this-is-paid-content-instead)

PR #191, opened 2017-05-23
Fix link snap branding

kelvin-chappell:
So that it no longer makes unnecessary calls to capi.


janua:
I wonder is it worth adding the same logic that the CAPI client has around containing a query parameter in a URL?
Actually, looking at it, `URI.getPath` is probably stripping out any possible query parameters anyway (Even though they may be removed upstream from facia-tool, not sure.)

👍

kelvin-chappell:
Yes, [this test](https://github.com/guardian/facia-scala-client/pull/191/files#diff-2774b6a33d5cbc00ca710e4397acfed9R177) shows that the query string is stripped out.

PR #193, opened 2017-05-24
Improve link snap branding

kelvin-chappell:
By reducing the number of pointless capi lookups and considering a failed lookup to be a success.


janua:
Great! 👍

PR #194, opened 2017-07-25
Update aws dependency

mchv:


janua:
Apart from proper version bump and the changelog. Looks all good! 👍

PR #195, opened 2017-09-06
Add support for play 2.6

TBonnin:
Because joda support has been removed from play-json 2.6
we are now providing our own implicit Formats for joda.Datetime to
ensure backward compatibility and cross play version compilation

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
@@ -3,6 +3,7 @@ package com.gu.facia.client.models
 import play.api.libs.json._
 import org.joda.time.DateTime
 import play.api.libs.json.{JsValue, Json}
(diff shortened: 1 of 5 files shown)


facia-json/src/main/scala/com/gu/facia/client/models/Collection.scala
```diff
@@ -3,6 +3,7 @@ package com.gu.facia.client.models
 import play.api.libs.json._
 import org.joda.time.DateTime
 import play.api.libs.json.{JsValue, Json}
+import com.gu.facia.client.json.JodaFormat._
```
mchv: Because you have imported `play-json-data`, you should be able to add 
```
import play.api.libs.json.JodaWrites._
import play.api.libs.json.JodaReads._
```
I think it provides the same as your custom one

mchv:
As discussed this is for cross-play-version compilation so the approach is very sensible!

PR #198, opened 2017-10-30
Pass in provided arguments in recursive call

regiskuckaertz:
Some editors picks are failing to be indexed in CAPI. We traced the origin back to the case where a collection is backfilled by its parent collection: in this case, the CAPI query does not request the required fields (`trailText`, etc.). Because of that, the generated JSON payload is incomplete.

Example [here](http://api.nextgen.guardianapps.co.uk/au/lifeandstyle/relationships/lite.json) (scroll down until you find two items with missing `trailText`)


regiskuckaertz:
It's not an isolated case:

screen shot 2017-10-30 at 13 52 48

janua:
:+1:

PR #204, opened 2018-03-07
Remove leading slash from capi id

tomrf1:
Users sometimes add capi backfill paths with a leading-slash, and facia-tool allows this. This breaks requests to capi preview because it creates urls with a double slash in the path, and this causes issues with api-gateway authorisation.

I'm fixing it here because it's used by both facia-press and mapi, and it's too late to fix it in facia-tool.


LATaylor-guardian:
👍

mchv:
Sounds sensible good 🐛 💀 @tomrf1 !

PR #206, opened 2018-04-11
Upgrade to CAPI Scala client v12

regiskuckaertz:
There has been several breaking changes in v12:

- the default http client has been moved into a separate project (`content-api-client-default`), users must either switch to that package or provide their own implementation (see [here](https://github.com/guardian/content-api-scala-client/blob/master/README.md))
- `ContentApiClientLogic` has become `ContentApiClient`
- utility functions have moved to the companion of that trait

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-#### 2.5.5
+#### 2.6
   
   - Upgrade Content API dependency to `12.0`
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+#### 2.5.5
+  
+  - Upgrade Content API dependency to `12.0`
+
 #### 2.4
   
   - Upgrade Content API dependency to `11.33`
--- a/fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
@@ -2,7 +2,7 @@ package com.gu.facia.api.contentapi
 
 import java.net.URI
 
-import com.gu.contentapi.client.ContentApiClientLogic
+import com.gu.contentapi.client.ContentApiClient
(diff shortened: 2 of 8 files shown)


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+#### 2.5.5
```
alexduf: My only comment is: I'm about to publish 2.5.5, I think you should go for 2.6.0 as you're pushing a contract change

regiskuckaertz:
Isn't there CI running for this project? I'd feel more at ease 😅

CHANGES.md
regiskuckaertz: Good catch!

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -24,21 +24,16 @@ object ContentApi extends StrictLogging {
   type AdjustSearchQuery = SearchQuery => SearchQuery
   type AdjustItemQuery = ItemQuery => ItemQuery
 
-  def buildHydrateQueries(client: ContentApiClientLogic, ids: List[String], adjustSearchQuery: AdjustSearchQuery = identity): Response[Seq[SearchQuery]] = {
-    def queryForIds(ids: Seq[String]) = adjustSearchQuery(client.search
-      .ids(ids mkString ",")
-      .pageSize(ids.size)
-      .showFields("internalPageCode"))
-
+  def buildHydrateQueries(client: ContentApiClient, ids: List[String], adjustSearchQuery: AdjustSearchQuery = identity, itemQueries: ItemQueries = ItemQueries): Response[Seq[SearchQuery]] = {
```
tomrf1: Why pass in `ItemQueries` as a parameter?

regiskuckaertz: Because it has to be mocked in the tests 😩

tomrf1:
👍 looks good

PR #208, opened 2018-08-23
Bump capi to 12.7

tomrf1:
The [change](https://github.com/guardian/content-api-scala-client/pull/262/files) to the `getResponse` method in content-api-client means that frontend needs this dependency to be updated at the same time


fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -34,7 +34,7 @@ object ContentApi extends StrictLogging {
   }
 
   def getHydrateResponse(client: ContentApiClient, searchQueries: Seq[SearchQuery])(implicit ec: ExecutionContext): Response[Seq[SearchResponse]] = {
-    Response.Async.Right(Future.traverse(searchQueries)(client.getResponse)) mapError { err =>
+    Response.Async.Right(Future.traverse(searchQueries)(client.getResponse(_))) mapError { err =>
```
regiskuckaertz: I think the addition of an implicit parameter group turned that function into a method.

PR #214, opened 2019-03-19
Update CAPI client

regiskuckaertz:
Updates to v12.18, which includes an [important fix to content-atom](https://github.com/guardian/content-atom/pull/128).

This brings several updates to the model, plus an important change where a transitive dependency with a wrong version leaked downstream.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -76,6 +76,7 @@ lazy val faciaJson = project.in(file("facia-json"))
     organization := "com.gu",
     name := "facia-json",
     resolvers ++= Seq(
+      Resolver.sonatypeRepo("releases"),
       Resolver.file("Local", file( Path.userHome.absolutePath + "/.ivy2/local"))(Resolver.ivyStylePatterns),
       "Typesafe Repository" at "http://repo.typesafe.com/typesafe/releases/"
     ),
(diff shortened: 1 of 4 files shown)


regiskuckaertz:
Perhaps this should be a minor version change, i.e. 2.7.0?

project/dependencies.scala
```diff
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "12.10"
+  val capiVersion = "12.18"
```
ajwl: So if I'm right - this is the key change here. And this new capiVersion of the libraries below now has the _old_ libthrift version that doesn't have a horrible backwards noncompatible breaking change?

build.sbt
```diff
@@ -76,6 +76,7 @@ lazy val faciaJson = project.in(file("facia-json"))
     organization := "com.gu",
     name := "facia-json",
     resolvers ++= Seq(
+      Resolver.sonatypeRepo("releases"),
```
ajwl: What does this do?

project/dependencies.scala
regiskuckaertz: Yes, precisely.

build.sbt
regiskuckaertz: When we publish libraries, they're available on Sonatype repos first, then on Maven. This adds the former, so we don't have to wait hours before seeing the new versions live.

project/dependencies.scala
ajwl: cool

ajwl:
👍

regiskuckaertz:
Thank you @ajwl

PR #219, opened 2019-07-31
Fixes #218: populates displayName from collectionConfig

### LINKED ISSUE #218: displayName used by Collection.fromCollectionJsonConfigAndContent seems wrong
Looking at the Australia Network Front at https://www.theguardian.com/au right now, you will see that the 2nd container (normally called 'Spotlight') is called **'Weekend'**:

image

That's correct - at the weekend (it's already the weekend in Australia!) the Fronts Editors manually update https://fronts.gutools.co.uk/editorial/config to set the 'Spotlight' container to have the Display Name of 'Weekend' - after the weekend is over, it's manually renamed back again. This is getting displayed correctly in DotCom.

However, in Ophan, where we use the `facia-scala-client` library to get fronts information, [we are displaying the wrong name](https://dashboard.ophan.co.uk/info?path=/commentisfree/2019/jul/26/the-untold-story-of-when-donald-trump-met-boris-johnson#how-did-we-promote-this-on-www-theguardian-com) for the container - we display 'Spotlight', _not_ 'Weekend':

image

The [name Ophan displays comes from](https://github.com/guardian/ophan/blob/2ccbcdb08a1f71fb47f5c86d1def120f2e23570a/trigr/app/fronts/service/PromotionPoller.scala#L54-L60) the `com.gu.facia.api.models.Collection#displayName` field - that field is populated inside the `fapi-client` code by `com.gu.facia.api.models.Collection#fromCollectionJsonConfigAndContent`:

https://github.com/guardian/facia-scala-client/blob/51ce5edfd2905d192a640cfb7e8b7880ee1a3a88/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala#L23-L26

If you look at that code, you'll see that `collectionJson.displayName`is being used in preference to `collectionConfig.displayName` - is this the wrong way round?!

image

cc @jonathonherbert @simone-smith

simone-smith:
@rtyley and I have discussed this with @Reettaphant, and have agreed that we should be checking the `collectionConfig` first for the `displayName` and `href`. It seems that neither of these fields are being written into collection content any more - they are only configured through collection config. 

cc @jonathonherbert

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -23,8 +23,8 @@ object Collection {
   def fromCollectionJsonConfigAndContent(collectionId: String, collectionJson: Option[CollectionJson], collectionConfig: CollectionConfig): Collection = {
     Collection(
       collectionId,
-      collectionJson.flatMap(_.displayName).orElse(collectionConfig.displayName).getOrElse("untitled"),
-      collectionJson.flatMap(_.href).orElse(collectionConfig.href),
+      collectionConfig.displayName.orElse(collectionJson.flatMap(_.displayName)).getOrElse("untitled"),
+      collectionConfig.href.orElse(collectionJson.flatMap(_.href)),
       collectionJson.map(_.live).getOrElse(Nil),
       collectionJson.flatMap(_.draft),
       collectionJson.flatMap(_.treats).getOrElse(Nil),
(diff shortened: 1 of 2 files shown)


jonathonherbert:
@rtyley @simone-smith couldn't be happier with this, thanks for investigating -- @Reettaphant thanks for confirming 👍 

I wonder how many other orphaned or redundant properties exist on collections; it should be fairly trivial to write a script to check? I'll add a card for health week.

philmcmahon:
Hope releasing this goes ok. I have done it before on my new laptop, happy to give it a go if you encounter difficulties

PR #222, opened 2019-09-05
Fix image override case: Change behaviour so fronts tool decides replacement image

ajwl:
# The problem 
A rare problem came up on the UK front page where the imageReplace was not being shown as the thumbnail for a story. The default CAPI image was being used instead. 

*Features of the story*: 
-> Opinion/comment piece 
-> It had no cutout image supplied or available from TagManager
-> An imageReplace had been applied in the fronts tool overriding the default image from CAPI.

NB: The standard format for an opinion piece on fronts is to use a cutout image / quote headline / show byline. 

*Two opinion pieces on a front - 1st with cutout, 2nd with standard image* 
Screenshot 2019-09-06 at 11 01 45
 
*Other aspects of the problem*
-> If a cutout image was added, even if it was not used, it solved the problem and image replace worked correctly. This is the normal case, as the vast majority of opinion pieces have a cutout available even if not used 
-> Fronts tool was correctly sending the data with the `imageReplace` property set to true and the new image supplied in the url. 
-> Frontend was correctly rendering what it received from facia-press.
-> Facia-press was correctly pressing the uk front. 
-> The same happens in v1, so this problem has obviously been around for a while, but only turns up infrequently because it applies to a small section of stories. 

## Suggested fix

From how I see it, the fronts tool should take the decision as to whether a cutout is used, rather than this being hardcoded as the default by this library. 

Therefore I am removing [line 62](https://github.com/guardian/facia-scala-client/pull/222/files#diff-28f7d5018617d6ae34f67050345e19e2R62) that hardcodes that for opinion/comment pieces. That should be set by [line 82](https://github.com/guardian/facia-scala-client/pull/222/files#diff-28f7d5018617d6ae34f67050345e19e2R82) taking it from the fronts tool trailMeta instead. 

This cedes control for setting the replaceImage and the cutoutImageReplace to the fronts tool metadata rather than guessing that an Opinion story will definitely have a cutout image. 

Setting it to true by default meant that it failed [this if statement](https://github.com/guardian/frontend/blob/88cfa609c73545085c3e5f3921631ec344a3eb83/common/app/model/ImageOverride.scala#L11) in facia press: 

```(scala)
val (maybeSrc, maybeWidth, maybeHeight) = image match {
      case cutout: Cutout => (Some(cutout.imageSrc), cutout.imageSrcHeight, cutout.imageSrcWidth)
      case replace: Replace => (Some(replace.imageSrc), Some(replace.imageSrcHeight), Some(replace.imageSrcWidth))
      case _ => (None, None, None)
    }
```

## Json produced by fronts

This image override worked because a cutout was present. 
```
"meta": {
"imageSrcWidth": "2625",
"headline": "It will wreck it: Boris Johnson’s electoral gamble risks wrecking the Tory party",
"imageSrcOrigin": "https://media.test.dev-gutools.co.uk/images/94b46e4078134a5b04a868c4054658a33348d5ec",
"imageSrcHeight": "1575",
"imageCutoutSrc": "https://uploads.guim.co.uk/2017/10/06/Jonathan-Freedland,-L.png",
"showByline": true,
"showQuotedHeadline": true,
"imageSrcThumb": "https://s3-eu-west-1.amazonaws.com/media-origin.test.dev-guim.co.uk/94b46e4078134a5b04a868c4054658a33348d5ec/0_104_2625_1575/500.jpg",
"imageSrc": "https://s3-eu-west-1.amazonaws.com/media-origin.test.dev-guim.co.uk/94b46e4078134a5b04a868c4054658a33348d5ec/0_104_2625_1575/master/2625.jpg",
"imageReplace": true,
"imageCutoutReplace": false,
"group": "1"
}
```

This one didn't because a cutout was not present:
```
"meta": {
"imageSrcWidth": "2595",
"headline": "It fell: Boris Johnson's blustering strategy has fallen at the first hurdle",
"imageSrcOrigin": "https://media.test.dev-gutools.co.uk/images/6d3a83433a65f38231d8455929ae5c6e47f55883",
"imageSrcHeight": "1558",
"showByline": true,
"showQuotedHeadline": true,
"imageSrcThumb": "https://s3-eu-west-1.amazonaws.com/media-origin.test.dev-guim.co.uk/6d3a83433a65f38231d8455929ae5c6e47f55883/0_87_2595_1558/500.jpg",
"imageSrc": "https://s3-eu-west-1.amazonaws.com/media-origin.test.dev-guim.co.uk/6d3a83433a65f38231d8455929ae5c6e47f55883/0_87_2595_1558/master/2595.jpg",
"imageReplace": true,
"group": "1"
}
```

## Other issues with this library 

Variations of this problem have come up before. cf this trello card - 
https://trello.com/c/nwUyy7Lv/67-content-based-settings-do-not-get-saved-into-database-when-creating-skeleton 
@aug24 @akash1810

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
@@ -59,8 +59,7 @@ object ResolvedMetaData {
       cardStyle match {
         case com.gu.facia.api.utils.Comment => Default.copy(
           showByline = true,
-          showQuotedHeadline = true,
-          imageCutoutReplace = true)
+          showQuotedHeadline = true)
         case _ if isCartoonForContent(content) => Default.copy(showByline = true)
         case _ if isVideoForContent(content) => Default.copy(showMainVideo = true)
         case _ => Default


fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
```diff
@@ -59,8 +59,7 @@ object ResolvedMetaData {
       cardStyle match {
         case com.gu.facia.api.utils.Comment => Default.copy(
           showByline = true,
```
akash1810: `showByline` being `true` makes me really nervous as the `byline` field is optional in [the model](https://github.com/guardian/flexible-model/blob/2c9268bbceb4212de480ee47158f4b08eb4e2682/src/main/thrift/content.thrift#L1165-L1166)

Given this scenario is rare, would taking a more holistic view (i.e. remove all default metadata logic from the shared library) be better? That is, move to a place where the tooling is the canonical source of all metadata.

ajwl: yeah. I agree, if we are removing styling choices from here, we should probably remove all styling choices. It is a weird place for them.

ajwl:
Have added a 2nd commit which is a more drastic removal of these default styling options. Including defaults setting `mainVideo` and `showByline`

fapi-client/src/main/scala/com/gu/facia/api/utils/ResolvedMetaData.scala
```diff
@@ -55,33 +55,22 @@ object ResolvedMetaData {
...
 
-  def fromContent(content: Content, cardStyle: CardStyle): ResolvedMetaData =
-      cardStyle match {
-        case com.gu.facia.api.utils.Comment => Default.copy(
-          showByline = true,
-          showQuotedHeadline = true)
-        case _ if isCartoonForContent(content) => Default.copy(showByline = true)
-        case _ if isVideoForContent(content) => Default.copy(showMainVideo = true)
-        case _ => Default
-      }
```
ajwl: This function adds the default styles, so it's not needed now

ajwl:
Bumping this as the problem has come up again

jonathonherbert:
@akash1810 @ajwl this code looks fine, but if we're removing this logic here, I feel like we should take care to make sure it exists elsewhere. For example, AFAIK `isCartoonForContent` isn't handled in the Fronts tool, and as a result when editors add cartoons they may find that the output on platform isn't what they expect (no byline).

Given that this is causing problems for editors now, would it be safer to remove logic piece by piece, at least in this case, to make sure we don't cause more problems than we solve? I feel like we should be applying these other defaults elsewhere (Fronts tool) in tandem with their removal here.

ajwl:
@jonathonherbert yeah sensible... I can roll back the 2nd commit tomorrow and just fix the bug in hand

ajwl:
@jonathonherbert @akash1810 - have reduced this PR to the original small scope fix. Lmk if you are happy

ajwl:
thanks!

PR #223, opened 2019-10-16
Geotargeting collections: add 'TargetedTerritory' to CollectionConfigJson 

jonathonherbert:
This PR adds an optional 'TargetedTerritory' property to the CollectionConfigJson class, allowing us to specify a targeted territory on a per-Collection basis.

It's needed to support the upcoming work targeting containers to specific geographic regions.

At the moment, we only support a single territory per Collection -- let me know if this should be a `List[TargetedTerritory]`!

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -76,6 +76,26 @@ object CollectionPlatform {
   }
 }
 
+sealed trait TargetedTerritory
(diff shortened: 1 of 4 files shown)


alexduf:
>At the moment, we only support a single territory per Collection -- let me know if this should be a List[TargetedTerritory]

I think it makes sense as a single value. If we were to have one container to span across multiple countries for instance, then we would probably adapt the geolocation in the first place, like how we're grouping eu-27.

jonathonherbert:
@alexduf brilliant, thanks for the feedback 👍

PR #225, opened 2019-10-21
Upgrade release pipeline

alexduf:
I haven't fully tested it yet

If no one disagree, I'd like to do another release after this PR is merged, just to confirm that it works.
So 3.0.8 and 3.0.9 would functionally be the same, but I'd prove the pipeline still works.

This came about as I was trying to release 3.0.8 this morning and couldn't explain why it wouldn't work so I decided to upgrade all the release libraries. 

(it turns out my issue was unrelated but we may as well upgrade all the libraries)

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,6 +1,6 @@
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.11")
+addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.12")
 
-addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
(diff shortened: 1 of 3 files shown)


project/plugins.sbt
```diff
@@ -1,6 +1,6 @@
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.11")
+addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.12")
```
jennygrahamjones: Is there a reason we don't use https://github.com/sbt/sbt-release ? It looks like this is the canonical version now

alexduf: I think it is already the case: https://github.com/sbt/sbt-release#usage (and I'm relieved because that would have been painful to migrate :) )

PR #226, opened 2019-10-21
Targeted Territory: update collection model with targetedTerritory

ajwl:
This property also needs to be added to the  `CollectionConfig` model.

CF: https://github.com/guardian/facia-scala-client/pull/223


ajwl:
@alexduf - need to amend model. Will let you know when we release

PR #231, opened 2019-11-19
update snaplink models with atomId property

ajwl:
When snap links are created in the Fronts Tool with links to atoms in the content api eg: 
`https://content.guardianapis.com/atom/interactive/interactives/2019/10/test-snap`. 

We want to extract the `atomId` and send it on in a new optional `atomId` property, now added to the Meta object. 

This makes it easier for downstream teams like Dotcom and MAPI to fetch the atom from CAPI. 

This is follow-up to this work to support Interactive Atoms in the Fronts Tool.  https://github.com/guardian/facia-tool/pull/1067


ajwl:
Thanks Alex. Have added a test fix. Will try to release now

PR #234, opened 2019-12-05
Add blockId to trail meta

jonathonherbert:
Adds an optional blockId property to our Trail objects.

Required to let [facia-tool](https://github.com/guardian/facia-tool) store block ids for breaking news notifications that refer to liveblogs, allowing notification clients to point the user to a specific block in a liveblog.


ajwl:
NB: adding AtomId required a few more changes. Are we likely to require some more of these? https://github.com/guardian/facia-scala-client/pull/231/files

jonathonherbert:
Great question, @ajwl -- going through those files,

- `curatedcontent.scala` -- we shouldn't have to worry about snaplinks, as this change only relates to articles
- `FaciaContentUtils.scala` -- there's nothing downstream that consumes the `atomId` yet (the notification client has its own model) and without a use case I was reluctant to add code, so we should be fine for now (although open to opinions here!)
- Test files -- because the change was very simple, I didn't add a test, although I'm happy to do this

I hope that makes sense, and we can always revisit if necessary.

ajwl:
👍 cool thanks for the update. Will proceed with this

PR #245, opened 2021-03-16
Handle theguardian.com short url correctly

buck06191:
## What does this change?

CAPI has been updated to use `www.theguardian.com` for short urls. I've added the appropriate regex here so that the facia client correctly removes this when getting the `discussionId` and added tests to check for this.

### Related PRs:
- https://github.com/guardian/frontend/pull/23565/files Temporary fix in Frontend that this supercedes
- https://github.com/guardian/content-api/pull/2547 Explanation for why I'm checking for `www.`
- https://github.com/guardian/frontend/pull/23337 Initial handling in Frontend
## How to test

`sbt test`

Correct discussion ID passed through from new short URLs.

## Have we considered potential risks?

The CAPI change has already occurred so this shouldn't be a breaking change. Even so, I've kept the handling of the old gu.com domain in.


fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
```diff
@@ -144,7 +144,9 @@ object FaciaContentUtils {
   )
   def maybeShortUrl(fc: FaciaContent) = fieldsGet(fc)(_.flatMap(_.shortUrl))
   def shortUrl(fc: FaciaContent): String = maybeShortUrl(fc).getOrElse("")
-  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(_.replaceFirst("^[a-zA-Z]+://gu.com", ""))
+  def shortUrlPath(fc: FaciaContent) = maybeShortUrl(fc).map(
+    _.replaceFirst("^https?://gu.com", "")
```
jfsoul, line 148: I don't think it should be necessary to keep the gu.com removal should it?

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
```diff
@@ -144,7 +144,9 @@ object FaciaContentUtils {
...
+      .replaceFirst("^https?://(www.)+theguardian.com", ""))
```
jfsoul, line 149: Might be unnecessary to make the `www.` optional as well? Plus do we need to escape any of the other characters i.e. `/` and `.`?

jfsoul (approved):

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
jfsoul: actually, sorry, you're not making it optional with `+`?

buck06191: Ah yeah this should have been `?` to make it optional. I guess it does make more sense to actually have it as a required part of the string given that this is parsing what is sent over from CAPI directly.

fapi-client/src/main/scala/com/gu/facia/api/utils/FaciaContentUtils.scala
buck06191: Yeah, I was keeping it in for safety, but I think that if CAPI has fully moved over to the longer domain now it should be safe to remove it.

Issue #249, opened 2021-06-17
Malformed URLs throw an error when parsing snaplinks


jonathonherbert:
## Problem

When the client encounters a malformed URL in a snap link, it [throws an error](https://github.com/guardian/facia-scala-client/blob/4c40e7c1764fcdc947cf7a7e4bfa4879809132d4/fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala#L119). This recently caused a problem where the accidental inclusion of a javascript bookmarklet in a collection caused a front to fail to press.

```scala
java.net.URISyntaxException: Illegal character in opaque part at index 22: javascript:(function(){document.body.appendChild(document.createElement('script')).src='https://dashboard.ophan.co.uk/assets/js/heatmap-bookmarklet.js';})();
	at java.net.URI$Parser.fail(URI.java:2847)
	at java.net.URI$Parser.checkChars(URI.java:3020)
	at java.net.URI$Parser.parse(URI.java:3057)
	at java.net.URI.<init>(URI.java:588)
	at com.gu.facia.api.contentapi.ContentApi$.toIdAndUri$1(ContentApi.scala:119)
	at com.gu.facia.api.contentapi.ContentApi$.$anonfun$linkSnapBrandingsByEdition$5(ContentApi.scala:130)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:273)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.map(TraversableLike.scala:273)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:266)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at com.gu.facia.api.contentapi.ContentApi$.linkSnapBrandingsByEdition(ContentApi.scala:130)
	at com.gu.facia.api.FAPI$.getLinkSnapBrandings(FAPI.scala:134)
	at com.gu.facia.api.FAPI$.getLiveLinkSnapBrandingsForCollection(FAPI.scala:120)
[...]

PR #250, opened 2021-06-18
Ignore snaps with bad links, rather than throwing

jonathonherbert:
## What does this change?

An attempt at a fix for #249. See that issue for details.

## Dev notes

Because the failure to parse the URL only occurs when we're attempting to recover the 'branding' for the snapLink (which I don't fully understand, but IIUC cannot possibly function with an invalid link!), I think we're safe to just move on and continue parsing the rest of the response. Hence we log this error, but convert to an option and continue – if anyone knows better, please say!

## How to test

- Do the automated tests pass, and are you satisfied that they're adequate?
- Is the logging on failure adequate?
- Import a local build or a snapshot of this library into a consumer, and attempt to press a snap link with an invalid link. Does the press now succeed?


twrichards (approved):

jonathonherbert:
Tested w/ @frankie297, looks good on MAPI 👍

PR #252, opened 2021-07-12
Deduce SpecialReport from CAPI client

davidfurey:
## What does this change?

We currently store a hardcoded list of tags which are `Special Reports` in multiple locations.  This PR removes some of the duplication by relying on the CAPI client to identify special reports.


buck06191 (approved):
LGTM. Better not to hardcode a bunch of hashes

PR #253, opened 2021-09-09
add podcast tag to config fields

dskamiotis:
## What does this change?

Within Fronts we currently have various options in tags to help define the type of article, or to associate an apt tag to an article.

In this PR we are adding a `Podcast` field to the tags so that users can select this as a type of article.

## How to test

This can be tested using publishLocal and then running the facia-tool locally to check for changes

## Images

| Before  | After |
| --- | --- |
| image| image|

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -1,6 +1,5 @@
 package com.gu.facia.client.models
 
-import com.sun.corba.se.spi.legacy.interceptor.UnknownType
 import com.typesafe.scalalogging.StrictLogging
 import play.api.libs.json._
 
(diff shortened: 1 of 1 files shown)


facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -1,6 +1,5 @@
 package com.gu.facia.client.models
 
-import com.sun.corba.se.spi.legacy.interceptor.UnknownType
```
jonathonherbert, line 3: To double check – this is something we found was unused, so we're tidying up?

jonathonherbert (approved):
One quick question re: a removed import, otherwise, great work! 👍

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
dskamiotis: I think this may have been added from a previous commit as there is no unknownType in this file.
I assume my editor may have deleted unused imports?

It is running locally without errors in the log.

shtukas (approved):

PR #255, opened 2021-09-30
Territory exhaustive match

davidfurey:
## What does this change?

When preparing to release this library I noticed a warning about a non-exhaustive pattern match.  This PR makes the match exhaustive, returning a JSON parse error instead of throwing an exception if asked to parse an unknown territory.


srbd (approved):
LGTM!

mchv:
@davidfurey I think another improvement could be to turn non-exhaustive pattern matching warnings to errors with something like `Xfatal-warnings` in this project. What do you think?

davidfurey:
@mchv - indeed, I've opened https://github.com/guardian/facia-scala-client/pull/259 with this change

I'd appreciate your thoughts on https://github.com/guardian/facia-scala-client/pull/258 which should avoid new territories breaking out-of-date clients in future.

PR #258, opened 2022-02-03
Add unsupported territory

davidfurey:
## What does this change?

We have previously encountered unexpected behaviour when new territories are added without updating all projects to use the latest version of the client.

Many consumers of this client do not care about the territories, but new territories mean old clients can no longer parse the config correctly.

This PR adds a new "Unsupported" territory that will be returned if the client cannot parse the a territory in the config.


mchv:
Yes, I think this is the right way for client consumer to force them to handle the case when the territory is not recognised rather than dealing with a `JsError` and it follow the pattern used with `UnknownMetadata`

davidfurey:
Thanks for looking at this, and nice to see that I'm (accidentally) following an established pattern.

jonathonherbert (approved):
LGTM, 💯  on test!

PR #259, opened 2022-02-03
Treat compiler warnings as errors

davidfurey:
## What does this change?

Enables the warnings-as-errors compiler flag, and fixes all warnings.

This is primarily to ensure that that non-exhaustive matches are considered an error.


mchv:
👍

PR #260, opened 2022-02-04
Fix scala 2.13 warnings

davidfurey:
## What does this change?

Fixes the build, which was broken by #259.  Treating compiler warnings as errors is a nice idea, but I only tested it on Scala 2.12. This project is set up to cross compile to Scala 2.13 too.  Scala 2.13 deprecates a number of features, which this PR resolves.


fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -58,7 +59,7 @@ object FAPI {
       )
       collectionConfigs = collectionConfigJsons.map(CollectionConfig.fromCollectionJson)
     } yield {
-      (collectionIds, collectionsJsons, collectionConfigs).zipped.toList.map { case (collectionId, collectionJson, collectionConfig) =>
+      collectionIds.lazyZip(collectionsJsons).lazyZip(collectionConfigs).toList.map { case (collectionId, collectionJson, collectionConfig) =>
```
rtyley, line 62: Just want to check, would it be possible to use `scala-collection-compat` rather than building your own shim?

It looks like `ws lazyZip xs lazyZip ys` [is supported](https://github.com/scala/scala-collection-compat/blob/b26d836a629654c3117c6350df196bf025f2b7c7/compat/src/test/scala/test/scala/collection/LazyZipTest.scala#L20-L27) by [`scala-collection-compat`](https://github.com/scala/scala-collection-compat)?

rtyley (commented):
Don't forget to fix the tests! I'm seeing errors like this when running `Test / compile`

```
error] /Users/roberto/guardian/facia-scala-client/fapi-client/src/test/scala/com/gu/facia/api/utils/ResolvedMetaDataTest.scala:167:9: symbol literal is deprecated; use Symbol("showMainVideo") instead
[error]         'showMainVideo (false))
[error]         ^
```

I think this repo might need some CI 😃

davidfurey:
> Don't forget to fix the tests! I'm seeing errors like this when running `Test / compile`
> 
> ```
> error] /Users/roberto/guardian/facia-scala-client/fapi-client/src/test/scala/com/gu/facia/api/utils/ResolvedMetaDataTest.scala:167:9: symbol literal is deprecated; use Symbol("showMainVideo") instead
> [error]         'showMainVideo (false))
> [error]         ^
> ```
> 
> I think this repo might need some CI 😃

Thanks, fixed.

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
davidfurey: Amazing, I looked for this but only found https://github.com/giabao/scala-collection-compat-lazyzip/ which didn't look maintained.

fapi-client/src/main/scala/com/gu/facia/api/contentapi/ContentApi.scala
```diff
@@ -99,7 +99,7 @@ object ContentApi extends StrictLogging {
   def parseQueryString(queryString: String): Seq[(String, String)] = {
     val KeyValuePair = """([^=]+)=(.*)""".r
 
-    queryString split "&" collect {
+    queryString.split("&").toIndexedSeq.collect {
```
rtyley, line 102: Can just do `toSeq` - don't think the performance characteristics of `IndexedSeq` are required?

```suggestion
    queryString.split("&").toSeq.collect {
```

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
```diff
@@ -92,7 +92,7 @@ class IntegrationTest extends FreeSpec with Matchers with ScalaFutures with Opti
 
     "will use the provided function to adjust the query used to hydrate content" ignore {
       val adjust: AdjustSearchQuery = q => q.showTags("tone")
-      val collection = collectionResponse.right.get
+      val collection = collectionResponse.toOption.get
```
rtyley, line 95: Not essential, just wanted to note that if you update to current ScalaTest (which would involve classname/imports updates), you can use the `EitherValues` trait and do `collectionResponse.value` rather than `.toOption.get`  (thanks to https://github.com/scalatest/scalatest/pull/1895).

rtyley (approved):
Awesome work man!

fapi-client/src/test/scala/com/gu/facia/api/integration/IntegrationTest.scala
davidfurey: Nice, this would be a good follow up improvement, but I'll leave it out since this is already a bigger PR than I'd hoped.

PR #265, opened 2022-03-21
Transform optional caption on facia image (for slideshow images only)

Fweddi:
## What does this change?

#264 added the caption to the slideshow asset. This PR ensures that when these assets are transformed, we handle the optional caption on the asset.

## How to test

A new test should pass, showing the caption is transformed when set. 

The edited tests should pass, showing that we can handle images without captions (both in a slideshow context, and in an 'image replace' context - where captions should never be set). This should ensure we are backwards compatible.

The captions should now be passed all the way through to the tooling as intended in #264.


jamesgorrie (approved):

OllysCoding (approved):
+1 this all make sense to me! Looks good :)

PR #281, opened 2022-10-26
adds specialReport palette to fronts

dskamiotis:
## What does this change?

In preparation for the release of a new type of special report a new Palette is required to facilitate this.
In this PR we are adding a `SpecialReportAlt` field to the tags so that users can select this as a type of article.

Name: SpecialReportAlt (TBC)

Tested Locally by `+publishLocal` and added to the lib in `facia-tool`. 

## How to test
Tested Locally by `+publishLocal` and added to the lib in `facia-tool` and check: https://fronts.local.dev-gutools.co.uk/v2

This can be tested using publishLocal and then running the facia-tool locally to check for changes

## Images

|Name Here|
|--|

## Accessibility

## Deployment

### DIFF
--- a/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
+++ b/facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
@@ -42,6 +42,7 @@ case object UnknownMetadata extends Metadata
 case object LongRunningAltPalette extends Metadata
 
 case object SombreAltPalette extends Metadata
+case object SpecialReportAlt extends Metadata
 
 object Metadata extends StrictLogging {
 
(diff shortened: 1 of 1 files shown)


facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -59,7 +60,8 @@ object Metadata extends StrictLogging {
     "EventAltPalette" -> EventAltPalette,
     "Podcast" -> Podcast,
     "LongRunningAltPalette" -> LongRunningAltPalette,
-    "SombreAltPalette" -> SombreAltPalette
+    "SombreAltPalette" -> SombreAltPalette,
+    "SpecialReportAlt" -> SpecialReportAlt
```
ioannakok, line 64: Can we please go with `SpecialReportAltPalette` for consistency?

dskamiotis: @ioannakok that sounds sensible will do 👍

ioannakok: Perfect, thank you very much!

davidfurey (approved):

PR #287, opened 2023-08-04
Add ETag-caching and AWS SDK v2 support

rtyley:
This change adds these improvements:

* Facia data is only re-downloaded & re-parsed if the S3 content has _changed_ (thanks to [ETag-caching](https://github.com/guardian/etag-caching))
* Independence from AWS SDK version (v1 vs v2) _(this PR can replace https://github.com/guardian/facia-scala-client/pull/286)_

The [ETag-caching](https://github.com/guardian/etag-caching) library itself is also being used in DotCom PROD, introduced with https://github.com/guardian/frontend/pull/26338.

### Usage

```scala
import com.gu.facia.client.{ApiClient, Environment}
import com.gu.etagcaching.aws.sdkv2.s3.S3ObjectFetching
import software.amazon.awssdk.services.s3.S3AsyncClient

val s3AsyncClient = S3AsyncClient.builder().region(...).credentialsProvider(...).build()

val apiClient = ApiClient.withCaching(
  "facia-tool-store",
  Environment.Prod,
  S3ObjectFetching.byteArraysWith(s3AsyncClient)
)
```

_PR using this updated version of the FAPI client: https://github.com/guardian/ophan/pull/6741_

### Independence from AWS SDK version (v1 vs v2)

Ideally, the whole of `facia-scala-client` would be independent of AWS SDK version - we'd _like_ consumers of this library to be able to use whatever AWS SDK version they want, without us pulling in dependency on either SDK version. 

For `facia-scala-client` this is an attainable goal, as the only AWS API action it performs is fetching from S3, and [guardian/etag-caching](https://github.com/guardian/etag-caching) provides the [`S3ByteArrayFetching`](https://github.com/guardian/etag-caching/blob/v6.0.0/aws-s3/base/src/main/scala/com/gu/etagcaching/aws/s3/package.scala#L6-L22) abstraction that encapsulates this action without tying to a specific AWS SDK version.

Due to legacy code compatibility, we can't completely remove AWS SDK v1 from `fapi-client` for now, but we _have_ removed it from `fapi-client-core`, which is the new home of `com.gu.facia.client.ApiClient`, which is now a trait, with 2 constructor methods that provide different implementations:

* **`ApiClient()`**  - legacy, using the existing `com.gu.facia.client.S3Client` abstraction on S3 behaviour
* **`ApiClient.withCaching()`** - provides ETag-based caching and is independent of AWS SDK version - the consumer just needs to provide an appropriate instance of `com.gu.etagcaching.aws.sdkv2.s3.S3ObjectFetching` (ie with `"com.gu.etag-caching" %% "aws-s3-sdk-v2"` and `S3ObjectFetching.byteArraysWith(s3AsyncClient)`, introduced with https://github.com/guardian/etag-caching/pull/65)

### Solved problems

* **Noisy logging associated with absent collection JSON** - the `etag-caching` library has been updated with https://github.com/guardian/etag-caching/pull/56 to avoid excessive logging that would occur in the Facia client, due to it typically trying to access collections that aren't yet persisted: https://github.com/guardian/facia-scala-client/pull/32.


facia-json/src/main/scala/com/gu/facia/client/AmazonSdkS3Client.scala
```diff
@@ -1,18 +1,14 @@
...
+
 import scala.concurrent.{ExecutionContext, Future, blocking}
 import scala.util.Try
 
-/** For mocking in tests, but also to allow someone to define a properly asynchronous S3 client. (The one in the AWS
-  * SDK is unfortunately synchronous only.)
-  */
-trait S3Client {
-  def get(bucket: String, path: String): Future[FaciaResult]
-}
```
rtyley, line 15: Note the `com.gu.facia.client.S3Client` trait still exists, it's just moved to the `fapi-client-core` artifact - couldn't delete it without badly breaking binary compatability for existing users!

rtyley:
As discussed at standup, I'm going to do a release of `facia-scala-client` as it currently is, and get that in use, before introducing this new update.

build.sbt
```diff
@@ -6,10 +6,10 @@ name := "facia-api-client"
 
 description := "Scala client for The Guardian's Facia JSON API"
 
-ThisBuild / scalaVersion := "2.13.14"
+ThisBuild / scalaVersion := "2.13.15"
 
 val sonatypeReleaseSettings = Seq(
-  releaseVersion := fromAggregatedAssessedCompatibilityWithLatestRelease().value,
+  // releaseVersion := fromAggregatedAssessedCompatibilityWithLatestRelease().value,
```
rtyley, line 12: We have to comment this out due to this issue, as we're introducing the new `fapi-client-core` artifact:

* https://github.com/guardian/gha-scala-library-release-workflow/issues/33

...we'll re-enable it immediately after the next release.

facia-json/src/main/scala/com/gu/facia/client/ApiClient.scala
```diff
@@ -1,37 +1,82 @@
...
+    case FaciaUnknownError(message) => throw BackendError(message)
   }
 
-  def config: Future[ConfigJson] =
-    retrieve[ConfigJson](s"$environment/frontsapi/config/config.json").map(_ getOrElse {
-      throw new BackendError("Config was missing!! OH MY GOD")
-    })
+  private def parseBytes[B: Format](bytes: Array[Byte]): B =
+    Json.fromJson[B](Json.parse(new String(bytes, Encoding))) getOrElse {
+      throw JsonDeserialisationError(s"Could not deserialize JSON")
```
fredex42, line 52: It might be nice to log out the actual error here - maybe one for a future pr? @rtyley

fredex42 (approved):
Looking great 😁

build.sbt
rtyley: Re-enabled with 42fa0bd1e44a59c4270f78fa305de7063eb0c044.

davidfurey: Sadly this created the impression that this release was much more compatible with the previous release than it actually is, making https://github.com/guardian/mobile-apps-api/pull/3503 a bigger piece of work than I was expecting.

rtyley: You're right, this PR (https://github.com/guardian/facia-scala-client/pull/287) was released as a patch release ([v14.0.1](https://github.com/guardian/facia-scala-client/releases/tag/v14.0.1)) when it should have gone out as a major release (because it _isn't_ binary or source compatible), bumping the first integer of the version number.

> making [guardian/mobile-apps-api#3503](https://github.com/guardian/mobile-apps-api/pull/3503) a bigger piece of work than I was expecting.

I've updated the release notes for v14.0.1 to list the significant incompatibilities found in this PR.

Issue #301, opened 2023-12-15
Runtime compatibility errors (eg `NoSuchMethodError`)


rtyley:
The `fapi-client` libraries depend on several Guardian libraries that can easily become incompatible with each other:

* `content-api-client`
* `content-api-models`
* ...

When app servers use `fapi-client` and (possibly independently) `content-api-client`/`content-api-models`, the ultimate versions of those libraries used can be binary-incompatible. This will lead to fatal runtime errors at unpredictable times.

### How unpredictable are these runtime errors?

Upgrading to `fapi-client` [v4.0.6](https://github.com/guardian/facia-scala-client/releases/tag/v4.0.6) led to fatal errors on **in production** on [fronts.gutools.co.uk](https://fronts.gutools.co.uk/) - not immediately on startup, but **when a user made a search query that _happened_ to return podcast results**. This is obviously very hard to anticipate and test for.

The upgrade that caused this issue was actually [content-api-models v17.5.1 → v17.5.2](https://github.com/guardian/content-api-models/compare/v17.5.1...v17.5.2) - a very innocent-looking change.

### What can we do?

See [_PREVENTING VERSION CONFLICTS WITH VERSIONSCHEME_](https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#versionscheme-librarydependencyschemes-and-sbt-150) - in summary, you need all 3 of these things:

* **Easy** : Use sbt 1.5.0 or higher everywhere (well, we're probably already doing that)
* **Easy** : Add `ThisBuild / versionScheme := Some("early-semver")` configuration to libraries we want to provide protection for (eg [facia-scala-client](https://github.com/guardian/facia-scala-client), [content-api-scala-client](https://github.com/guardian/content-api-scala-client), [content-api-models](https://github.com/guardian/content-api-models), etc)
* **Hard** : ensure that library updates are released with correctly incremented `major.minor.patch` version number updates, that respect [Scala's Recommended Versioning Scheme](https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#versioning-scheme---communicating-compatibility-breakages), which is an extended set of rules adding on to plain [semver](https://semver.org/).

This last part is being addressed with https://github.com/guardian/gha-scala-library-release-workflow and `sbt-version-policy`.

davidfurey:
@roberto https://github.com/guardian/facia-scala-client/pull/305 suggests that this isn't required, can we close this issue?

[...]

PR #305, opened 2024-01-25
Prevent runtime incompatibility with 'early-semver' `content-api-scala-client`

rtyley:
Now that both `content-api-models` and `content-api-scala-client` are both using 'early-semver' and adhering to it (thanks to [`gha-scala-library-release-workflow`](https://github.com/guardian/gha-scala-library-release-workflow) and `sbt-version-policy`) this should mean that it is **no longer possible** for a single project that depends on `content-api-scala-client`, `content-api-models` and `facia-scala-client` to have _incompatible_ versions of those artifacts.

This should prevent **horrible runtime compatibility errors** like https://github.com/guardian/facia-scala-client/issues/301, which occurred with the rollout of the innocent-looking changes in `facia-scala-client` [v4.0.6](https://github.com/guardian/facia-scala-client/releases/tag/v4.0.6).

# `sbt` messaging when incompatibility is introduced 

As ana example, this is the _"suspected to be binary incompatible"_ compile time error you'll get in [`facia-tool`](https://github.com/guardian/facia-tool/) if you try to upgrade the FAPI client library to 5.0.0, without also upgrading the version of the CAPI client to match:

```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error] 
[error] 	* com.gu:content-api-client_2.13:20.0.2 (early-semver) is selected over 19.2.1
[error] 	    +- com.gu:fapi-client-play28_2.13:5.0.0               (depends on 20.0.2)
[error] 	    +- com.gu:content-api-client-default_2.13:19.2.1      (depends on 19.2.1)
```

Just upgrading the CAPI client version to the ultimately-selected version (`20.0.2`) will fix the problem here.

In general, you want to upgrade any related library to be compiled against the _selected_ library version - so in the example above, any library we're using in `facia-tool` that depends on the CAPI client, needs to have been compiled against CAPI client version `20.0.x` (the `x` there is because we can tolerate PATCH level version differences).

### DIFF
--- a/fapi-client/src/test/scala/com/gu/facia/api/TestModel.scala
+++ b/fapi-client/src/test/scala/com/gu/facia/api/TestModel.scala
@@ -185,6 +185,7 @@ object TestModel {
     tags: Seq[StubTag],
     elements: Option[Seq[StubElement]]
   ) extends Content {
+    val channels: Option[Seq[ContentChannel]] = None // alternatively, create StubContentChannel
     def `type`: ContentType = ContentType.list.find(byName(typeName)(_)).get
     def sectionId: Option[String] = None
     def sectionName: Option[String] = None
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "19.2.3"
+  val capiVersion = "20.0.2"
 
   val awsSdk = "com.amazonaws" % "aws-java-sdk-s3" % "1.12.524"
   val commonsIo = "org.apache.commons" % "commons-io" % "1.3.2"


fapi-client/src/test/scala/com/gu/facia/api/TestModel.scala
```diff
@@ -185,6 +185,7 @@ object TestModel {
     tags: Seq[StubTag],
     elements: Option[Seq[StubElement]]
   ) extends Content {
+    val channels: Option[Seq[ContentChannel]] = None // alternatively, create StubContentChannel
```
rtyley, line 188: This is to cope with the channels model changes introduced in https://github.com/guardian/content-api-models/pull/211, which was released with `content-api-models` [v17.6.0](https://github.com/guardian/content-api-models/releases/tag/v17.6.0) in April 2023.

As none of the tests in `facia-scala-client` seem to care about the value of `channels`, its easiest to just make this value `None` rather than create a `StubContentChannel` - but if in the future tests do eventually need this value, we can make one.

project/dependencies.scala
```diff
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "19.2.3"
+  val capiVersion = "20.0.2-PREVIEW.update-to-latest-version-of-content-api-models-scala.2024-01-25T1416.253f60e5"
```
rtyley, line 4: This is the preview of https://github.com/guardian/content-api-scala-client/pull/405

PR #306, opened 2024-02-23
Bump CAPI scala client

davidfurey:
## What does this change?

Upgrades to the newest CAPI scala client. So that projects that depend on CAPI scala client and facia-scala-client can upgrade to the newest CAPI client which is binary incompatible with the current version used by facia-scala-client


simonbyford (approved):
LGTM!

PR #307, opened 2024-02-28
Add Play v2.9 support to aid Facia Tool upgrade

rtyley:
Whether you get benefit from going 2.8→2.9→3.0 (over just going 2.8→3.0) is probably a factor of how big and complicated the project being upgraded is (see also https://github.com/guardian/maintaining-scala-projects/issues/4) - I'd guess Facia Tool is quite big, so breaking the changes down into smaller chunks might be worth it in terms of avoiding risk.

Obviously we're encouraging people to get onto Play 3.0 to avoid the Akka BSL licensing issue!

Note also that Play 2.9 actually uses Play-Json *2.10*, which is a bit surprising!

### DIFF
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "21.0.0"
+  val capiVersion = "22.0.0"
 
   val awsSdk = "com.amazonaws" % "aws-java-sdk-s3" % "1.12.524"
   val commonsIo = "org.apache.commons" % "commons-io" % "1.3.2"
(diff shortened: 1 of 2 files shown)


project/dependencies.scala
```diff
@@ -27,6 +27,7 @@ object Dependencies {
   object PlayJsonVersion {
     val V27 = PlayJsonVersion("27", "com.typesafe.play", "2.7.4")
     val V28 = PlayJsonVersion("28", "com.typesafe.play", "2.8.2")
+    val V29 = PlayJsonVersion("29", "com.typesafe.play", "2.10.4", supportsScala3 = true)
```
rtyley, line 30: The `supportsScala3` flag isn't actually used here - but it doesn't mean it _requires_ Scala 3! Just that, in principle, we could create Scala 3 artifacts for this version.

Fweddi (approved):

PR #309, opened 2024-03-06
Update aws-java-sdk-s3 to remove ion-java

emdash-ie:
There’s a vulnerability in ion-java, which is depended on by the current version of aws-java-sdk-s3, but removed from 1.12.638 forward. This change updates that dependency to its newest version to remove the ion-java dependency and facilitate removal of the vulnerability from projects which depend on this one.

## Deployment

### DIFF
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -3,7 +3,7 @@ import sbt._
 object Dependencies {
   val capiVersion = "23.0.0"
 
-  val awsSdk = "com.amazonaws" % "aws-java-sdk-s3" % "1.12.524"
+  val awsSdk = "com.amazonaws" % "aws-java-sdk-s3" % "1.12.673"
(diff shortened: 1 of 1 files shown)


Divs-B:
> This supersedes [PR 297](https://github.com/guardian/facia-scala-client/pull/297), which is a scala steward PR updating to a slightly less new version. Maybe we should just go with the scala steward PR?

Yes I think lets go ahead with Scala-steward PR(s), good to have upgrade anyway across the board,
lets see if that fixes ion-java too if not then will take ahead your PR.

Divs-B (approved):
@rtyley and I have tried this out on `apple-news` code and it is working as expected and we are able to fix `ion-java` based snyk high vuln. 
https://github.com/guardian/apple-news/pull/317#issuecomment-1983396589

PR #312, opened 2024-03-27
Bump and test content-api-client preview release version

Divs-B:
We are testing `content-entity` latest release that is out after adopting gha-scala-release-process init, 
the version is updated in `content-api-model` and `content-api-client` so we need to test these libraries preview releases in fapi client too.

The intention is to test all these preview releases in any of client consuming these linbraries, we have `apple-news` and ` pubflow` for example.

## How to test

Make a preview release. test in `apple-news`

Application should compile with no version conflicts between libraries
Application shoud run well without any erros.

### DIFF
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "26.0.0"
+  val capiVersion = "27.0.0"
 
   val awsSdk = "com.amazonaws" % "aws-java-sdk-s3" % "1.12.673"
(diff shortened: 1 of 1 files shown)


Divs-B:
Bumped production release and tested locally via `sbt test` . tests are passing.

rtyley (approved):

davidfurey (approved):

Divs-B:
Thanks @davidfurey and @rtyley  for approving. I will merge and make prod release.
Then we will be all done to use all in apple-news or pubflow or any client using them.

PR #313, opened 2024-04-03
Update CAPI Scala Client -> 26.0.0

Fweddi:
## What does this change?

This brings in the timeline element changes (https://github.com/guardian/content-api-scala-client/pull/414) - see also https://github.com/guardian/content-api-models/pull/245.

### DIFF
--- a/project/dependencies.scala
+++ b/project/dependencies.scala
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
-  val capiVersion = "25.0.0"
+  val capiVersion = "26.0.0"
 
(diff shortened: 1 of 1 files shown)


simonbyford (commented):
Excuse the dumb question but what are the implications of this change? Has/can it be tested?

alinaboghiu:
For the facia client there should be no implications as fronts will never directly use this new element, but because it's a major version change we need to update the fapi client to prevent dependency version conflicts in places like Frontend that use both libraries. I think.

alinaboghiu (approved):

PR #316, opened 2024-07-05
Update default collection type

cemms1:
## What does this change?

Swaps "fixed/small/slow-**VI**" for "fixed/small/slow-**IV**"

## Why

The container type "fixed/small/slow-VI" does not exist, so it looks like this might be a mistake. MAPI updated the equivalent logic in 2017: https://github.com/guardian/mobile-apps-api/pull/1038

Though, I'm not even sure this is ever used! The git blame says this was last updated about 10 years ago 😵 (https://github.com/guardian/facia-scala-client/pull/31)

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
@@ -37,7 +37,7 @@ case class CollectionConfig(
     frontsToolSettings: Option[FrontsToolSettings])
 
 object CollectionConfig {
-  val DefaultCollectionType = "fixed/small/slow-VI"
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):
Yeah, wonder what happened? Looks like it may have previously been a valid container type (at least looking at the [list of containers](https://github.com/guardian/facia-scala-client/commit/a890773b2d02e9faa04ea934d9c1148088447396#diff-62f3010f9a4d641edfbf2a3f1b98abeec9e0a4456991d4891c70a6db646c6319R70) in the config when this was originally written), but clearly isn't a thing now... And we probably do want the `DefaultCollectionType` to be a valid collection!

PR #318, opened 2024-07-24
Drop Scala 2.12 and Play 2.7, begin partial support for Scala 3

rtyley:
Fixes #317, meaning that PRs like #287 can use Scala 2.13 features. Also dropping Play 2.7- [Play 2.8 was end-of-life (EOL) on 31st May 2024](https://github.com/playframework/playframework/releases/tag/2.8.21), so dropping support for Play 2.7 seems very reasonable!

Running the tests under Scala 3 required 3690941515d0f0c3e97781f202d0bd1aa915e0a8, fixing a test so that it doesn't rely on JSON object's field order.

## Consumers of `facia-scala-client`

All 7 consumers of the `facia-scala-client` library are now on Scala 2.13 (last one to upgrade was https://github.com/guardian/story-packages/pull/195!), and Play 2.8 or above, so they do not pose blockers to dropping Scala 2.12 and Play 2.7:

- [guardian/ophan](https://github.com/guardian/ophan/blob/4d51e80d85b0dacae2bc6966ed6f2e380119f26e/build.sbt#L92) - [Play 3.0.4](https://github.com/guardian/ophan/blob/4d51e80d85b0dacae2bc6966ed6f2e380119f26e/project/plugins.sbt#L3)
- [guardian/frontend](https://github.com/guardian/frontend/blob/1a25c1d9e5d1787660cb0502a65252b424c22ed2/project/ProjectSettings.scala#L35) - [Play 3.0.4](https://github.com/guardian/frontend/blob/1a25c1d9e5d1787660cb0502a65252b424c22ed2/project/plugins.sbt#L12)
- [guardian/apple-news](https://github.com/guardian/apple-news/blob/30fa9ea149c9b8296bd58a196030b9012958bc07/build.sbt#L8) - [Play 3.0.2](https://github.com/guardian/apple-news/blob/30fa9ea149c9b8296bd58a196030b9012958bc07/project/plugins.sbt#L7)
- [guardian/editors-picks-uploader](https://github.com/guardian/editors-picks-uploader/blob/bfc5c3524751c6c51be3cdf8c6cab159b295dc45/build.sbt#L4) - AWS Lambda, doesn't use Play
- [guardian/facia-tool](https://github.com/guardian/facia-tool/blob/00b5de0c43ef0a9f474844def545e6295f7d632a/build.sbt#L11) - [Play 3.0.2](https://github.com/guardian/facia-tool/blob/00b5de0c43ef0a9f474844def545e6295f7d632a/project/plugins.sbt#L15)
[...]

PR #320, opened 2024-08-02
Fix release so that it does not attempt to use Scala 2.12

rtyley:
After https://github.com/guardian/facia-scala-client/pull/318 dropped support for Scala 2.12, release 8.0.1 correctly no longer published Scala 2.12 artifacts, but unfortunately the 'root' project (which does not have any code!) was still defaulting to using Scala 2.12, because that is the default sbt uses if no `scalaVersion` is set.

The next time we tried to do a release (https://github.com/guardian/facia-scala-client/pull/319#issuecomment-2263535032) the automated-compatibility checking failed, because it saw that one project was using Scala 2.12, and so it attempted to download all the non-existent Scala 2.12 artifacts for release 8.0.1[^1] - and failed.

To fix this, we can simply ensure that all modules in this sbt build default to Scala 2.13, not 2.12 - ie particularly, the 'root' module!

## Testing

We [ran a preview release on this PR](https://github.com/guardian/facia-scala-client/actions/runs/10213913578), and the automated-compatibility checking succeeded 👍 :

image

[^1]: This behaviour seems like a bug in `sbt-version-policy`?! `sbt-version-policy` is already aware of the `publish / skip := true` flag, so should know that `root` does not have any artifacts...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,6 +6,8 @@ name := "facia-api-client"
 
 description := "Scala client for The Guardian's Facia JSON API"
 
+ThisBuild / scalaVersion := "2.13.14"
+
 val sonatypeReleaseSettings = Seq(
   releaseVersion := fromAggregatedAssessedCompatibilityWithLatestRelease().value,
   releaseCrossBuild := true, // true if you cross-build the project for multiple Scala versions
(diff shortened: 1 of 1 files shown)


waisingyiu (approved):
👍

PR #344, opened 2025-02-19
Add deeply-read as a potential source of backfill content

davidfurey:
## What does this change?

Updates content-api-client to a version that knows about deeplyRead.

Adds Deeply Read articles (if present in CAPI response) to the backfilled content, in the same way as mostViewed is.

Hopefully the order of mostViewed/deeplyRead doesn't matter - I don't anticipate backfill queries searching for both mostViewed and deeplyRead


JustinPinner (approved):
CAPI client v34.0.0 is now available. This is 👍 from CAPI's perspective but should also get a 👍 from Curation and Fronts folks before merging.

Georges-GNM (approved):
Same approach as most viewed should work, so looks good to me

PR #345, opened 2025-02-20
Add GroupsConfig field

Fweddi:
## What does this change?

Extends the behaviour of Groups to include a Max Items field per Group. This enables us to set the desired card numbers for different sections of the container (Splash, Large, Small).

Allows us to be backwards compatible with old Groups field. In the future we can remove the Groups field entirely, which will help us save on the Case Class limit (see 1ef79548cfe6c12aac072e2651d679d4b975b827).

## How to test

Compiles and looks sensible.


facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
```diff
@@ -204,6 +204,12 @@ object DisplayHintsJson {
 
 case class DisplayHintsJson(maxItemsToDisplay: Option[Int], suppressImages: Option[Boolean] = None)
 
+object GroupsConfigJson {
+  implicit val jsonFormat: OFormat[GroupsConfigJson] = Json.format[GroupsConfigJson]
+}
+
+case class GroupsConfigJson(name: String, maxItems: Option[Int])
```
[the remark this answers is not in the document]
davidfurey: I don't know - the field will be used as a hard limit so max seems accurate?

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -192,18 +191,4 @@ object Collection {
...
-case class Group(get: Int)
-
-object Group {
-  def fromGroups(groups: Groups): List[Group] = {
-    groups.groups.collect {
-      case IntegerString(n) => Group(n)
-    }
-  }
-
-  def toGroups(groups: List[Group]): Groups = {
```
davidfurey, line 206: Was toGroups never used?

davidfurey:
I'd suggest either using `publishLocal` or creating a `preview` release of this change so that you can see how this works in practice in facia-tool before we actually merge+release this

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
Fweddi: It's a good point, I had the same thought.

facia-json/src/main/scala/com/gu/facia/client/models/Config.scala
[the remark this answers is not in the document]
Fweddi: So long as containers are allowed to include less than `maxItems` then this is correct. And that sounds like what we're doing.

Fweddi:
Fronts Tool compiles with a deprecation warning:
image
image

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
Fweddi: Looks like `fromGroups` is used (https://github.com/search?q=org%3Aguardian+fromGroups&type=code), but `toGroups` isn't (https://github.com/search?q=org%3Aguardian+toGroups&type=code).

Fweddi: However I'd be inclined to have `toGroups` and `toGroupsJson` for feature parity in case anyone needs them? (less need for `toGroups` since it's being deprecated soon).

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
```diff
@@ -22,7 +46,7 @@ case class CollectionConfig(
     collectionType: String,
     href: Option[String],
     description: Option[String],
-    groups: Option[Groups],
+    groupsConfig: Option[GroupsConfig],
```
davidfurey, line 49:
```suggestion
    groups: Option[GroupsConfig],
```

davidfurey: Nope, changed my mind - groupsConfig is fine

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
```diff
@@ -74,7 +98,7 @@ object CollectionConfig {
       collectionJson.collectionType getOrElse DefaultCollectionType,
       collectionJson.href,
       collectionJson.description,
-      collectionJson.groups.map(Groups),
+      collectionJson.groupsConfig.map(GroupsConfig.fromGroupsConfigJson).orElse(collectionJson.groups.map(GroupsConfig.fromGroups)),
```
davidfurey, line 101: Whitespace seems strange here?

davidfurey (approved):
One non-blocking comment

fapi-client/src/main/scala/com/gu/facia/api/models/collectionconfig.scala
Fweddi: I think I fixed this 👍

PR #346, opened 2025-02-25
Add method to apply default boost levels

davidfurey:
## What does this change?

Adds a new method applyDefaultBoostLevelsAndGroups which allows platforms to apply the correct boosts and groups to each card after backfilling and deduping has been completed.

It will also ensure that groups within containers are correctly limited to their max items and have appropriate card sizes for flexible/general.

There a two versions of the new method. One is very generic, which allows dotcom to use it with its own internal card data structure. One is tied to the `FaciaContent` type, which is convenient for MAPI.

## Deployment

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
@@ -5,10 +5,11 @@ import com.gu.contentapi.client.model.v1.Content
 import com.gu.facia.api.contentapi.ContentApi.{AdjustItemQuery, AdjustSearchQuery}
 import com.gu.facia.api.contentapi.{ContentApi, LatestSnapsRequest, LinkSnapsRequest}
 import com.gu.facia.api.models._
-import com.gu.facia.api.utils.BackfillResolver
+import com.gu.facia.api.utils.{BackfillResolver, BoostLevel, ContentProperties}
 import com.gu.facia.client.ApiClient
-import scala.collection.compat._
 
+import scala.PartialFunction.condOpt
+import scala.collection.compat._
 import scala.concurrent.{ExecutionContext, Future}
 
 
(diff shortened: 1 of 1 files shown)


fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -189,4 +190,87 @@ object FAPI {
...
+    setBoostLevel: (T, BoostLevel) => T
+  ): T =
+    if (allowedBoosts.contains(getBoostLevel(content)))
+      content
+    else
+      setBoostLevel(content, default)
+
+  case class GroupBoostConfig(default: BoostLevel, allowedBoosts: List[BoostLevel], maxItems: Int, group: String)
+
+  private val flexibleGeneralBoosts = List(
```
Georges-GNM, line 222: I think I remember these were the agreed defaults for backfilled content, but just wanted to double check - so 1 item in the splash, nothing in the very big or big groups, and 8 stories in standard?

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -189,4 +190,87 @@ object FAPI {
...
+      collectionType = collectionConfig.collectionType,
+      contents = contents,
+      getBoostLevel = _.properties.boostLevel,
+      setBoostLevel = (content, default) => updateContentProperties(content, content.properties.copy(boostLevel = default)),
+      setGroup = updateGroup
+    )
+  }
+
+  // NB - this will not behave as expected if there are "gaps" in the curated content (ie there are no curated cards in the
+  // second group but there are curated cards in the third group). But the fronts tool is supposed to prevent gaps.
```
Georges-GNM, line 254: I'm not sure we've got this implemented at the moment so maybe something we need to add there? As in it's possible to have a flexible general with nothing in a group - is this the kind of situation that would lead to unexpected behaviour? 
image

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
```diff
@@ -189,4 +190,87 @@ object FAPI {
...
+    // Very large
+    GroupBoostConfig(BoostLevel.MegaBoost, List(BoostLevel.MegaBoost), maxItems = 0, group = "2"),
+    // Large
+    GroupBoostConfig(BoostLevel.Boost, List(BoostLevel.Boost, BoostLevel.MegaBoost), maxItems = 0, group = "1"),
+    // Standard
+    GroupBoostConfig(BoostLevel.Default, List(BoostLevel.Default, BoostLevel.Boost, BoostLevel.MegaBoost), maxItems = 8, group = "0")
+  )
+
+  private def boostsConfigFor(collectionType: String, groupsConfig: List[GroupConfig]): Option[List[GroupBoostConfig]] = {
+    val groupsConfigBiggestFirst = groupsConfig.reverse
```
Georges-GNM, line 234: `BiggestFirst` feels ambiguous as aside from the size of the item (e.g. splash) it made me think of the number of items in the group... Maybe something alluding to the prominence of the stories like BiggestDisplayFirst?

Georges-GNM (approved):
Few small comments but nothing blocking really - this looks good to me

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
davidfurey: These match up with https://github.com/guardian/facia-tool/blob/33614ebfe9d2ccda4d4dc70e7a8c0a9f4dce1b70/public/src/js/constants/defaults.js#L126 but in practice the values here will never actually be used, since the config will always include maxItems which override these defaults

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
davidfurey: `groupsConfigTopGroupFirst`?

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
Georges-GNM: Ah yeah, of course!

fapi-client/src/main/scala/com/gu/facia/api/FAPI.scala
davidfurey: I've discussed with Jonathan, and while it isn't ideal that the fronts tool doesn't bump things up, it is preferable that we get this change out in the platforms to support backfill

PR #347, opened 2025-02-26
Add max limit of sublinks shown a card

abeddow91:
## What does this change?
Adds a maximum number of sublinks that a card can have, dependent on its boost level. 

The limits for boost levels are as follows:

Default level => max 2 sublinks
Boost level => max 2 sublinks
Megaboost level => max 4 sublinks
Gigaboost level => max 4 sublinks

Not all cards have boost levels - only cards in beta containers do. For those that don't, we default to maximum 4

### Fronts tool showing excessive sublinks

|   Fronts tool    |
|-------------|
| ![tool][] |

### DCR limiting sublinks

| default & boost      | megaboost      | splash  / gigaboost    
|------------|-------------|------------|
| ![boost][] | ![megaboost][] | ![splash][] |

[tool]: https://github.com/user-attachments/assets/d690351c-cfe6-427b-8148-09ef8dc2cd8f
[splash]: https://github.com/user-attachments/assets/3d37b0e5-3c42-4cf5-9c74-2670f3173bde
[boost]: https://github.com/user-attachments/assets/3a1d2845-f7b2-4cbe-83ad-b1aa066b884d
[megaboost]: https://github.com/user-attachments/assets/63f1b799-5c5a-48e2-bd3c-38901cc27757

## Accessibility

## Deployment

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -40,7 +40,7 @@ object Collection {
   }
 
   /** Cards can have varying amounts of supporting content (sublinks) depending on their boostlevel**/
-  def maxSupportingItems(value: String): Int = value match {
+  private def maxSupportingItems(value: String): Int = value match {
     case BoostLevel.Default.label => 2
     case BoostLevel.Boost.label => 2
     case BoostLevel.MegaBoost.label => 4
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -56,9 +56,8 @@ object Collection {
     // if content is not in the set it was most likely filtered out by the CAPI query, so exclude it
     // note that this does not currently deal with e.g. snaps
     def resolveTrail(trail: Trail): Option[FaciaContent] = {
-         val boostLevel = trail.safeMeta.boostLevel
-         val maxItems = maxSupportingItems(boostLevel.getOrElse(""))
-
+      val boostLevel = trail.safeMeta.boostLevel
+      val maxItems = maxSupportingItems(boostLevel.getOrElse(""))
 
       content.find { c =>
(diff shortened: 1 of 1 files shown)


fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -44,13 +56,17 @@ object Collection {
     // if content is not in the set it was most likely filtered out by the CAPI query, so exclude it
     // note that this does not currently deal with e.g. snaps
     def resolveTrail(trail: Trail): Option[FaciaContent] = {
+         val boostLevel = trail.safeMeta.boostLevel
+         val maxItems = maxSupportingItems(boostLevel.getOrElse(""))
+
+
```
davidfurey, line 62:
```suggestion
      val boostLevel = trail.safeMeta.boostLevel
      val maxItems = maxSupportingItems(boostLevel.getOrElse(""))

```

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -36,6 +39,15 @@ object Collection {
       collectionConfig.targetedTerritory)
   }
 
+  /** Cards can have varying amounts of supporting content (sublinks) depending on their boostlevel**/
+  def maxSupportingItems(value: String): Int = value match {
```
davidfurey, line 43:
```suggestion
  private def maxSupportingItems(value: String): Int = value match {
```

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -36,6 +39,15 @@ object Collection {
...
+    case BoostLevel.Default.label => 2
+    case BoostLevel.Boost.label => 2
+    case BoostLevel.MegaBoost.label => 4
+    case BoostLevel.GigaBoost.label => 4
+    case _ => 4
```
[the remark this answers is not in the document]
abeddow91: Yep that's the intention.

The platforms currently apply a limit based on card type for the older cards (which is never higher than four). My thinking was that this allows us to handle fairground and non-fairground card context, whilst the platforms continue to apply a further limit where required for non-fairground cards

davidfurey (approved):

PR #355, opened 2025-04-14
Add fields for replacement video atom

Fweddi:
## What does this change?
Adds a `videoReplace` attribute to the properties object on FaciaContent. It also adds support for enriching FaciaContent with `atomId` and `mediaAtomData`.

### Details
Introduces an enrichment step during content retrieval to extract and attach the relevant media atom.

- If the content has `showMainVideo`, the main media atom is retrieved directly from the content and filtered out if expired.
- If the content has `videoReplace` set to true, the replacement video atom is retrieved via a CAPI query and also filtered if expired.
- Both approaches return `Atom.Media Data` if valid.

Since this step introduces a CAPI query, the return type of the content retrieval function has been updated to Response[Option[FaciaContent]].

`atomId` is a pre-existing field used to enrich Interactive Atoms. We have repurposed it to store an ID for a replacement Media (i.e. Video) Atom. In practice there should be no overlap between the two - video enrichment should only happen if the `videoReplace` field is set to true, and this should only be set if the atom is a Media atom.

If an atom cannot be found or is expired, the error is logged but swallowed, returning a None wrapped in a successful future. This ensured fronts still renderer if an atom cannot be retrieved. 

NB: The logic to filter out expired atoms was already present on both platforms. It may be preferable to always pass the atom to the platforms—regardless of expiration—and let them handle filtering during rendering. This would require input from Editorial and has been excluded from this PR to avoid scope creep.

This PR contains file formatting so it is advised to review with whitespace hidden to help reduce noise.

## How to test

Use a preview version of this release in MAPI or Frontend and validate that a replacement video is piped through to the end response.

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -1,10 +1,15 @@
 package com.gu.facia.api.models
 
-import com.gu.contentapi.client.model.v1.Content
-import com.gu.facia.api.contentapi.{LatestSnapsRequest, LinkSnapsRequest}
+import com.gu.contentapi.client.ContentApiClient
+import com.gu.contentapi.client.model.v1.{Content, ItemResponse}
+import com.gu.contentatom.thrift.AtomData
+import com.gu.contentatom.thrift.atom.media.MediaAtom
+import com.gu.facia.api.contentapi.{ItemQueries, LatestSnapsRequest, LinkSnapsRequest}
 import com.gu.facia.client.models.{CollectionJson, SupportingItem, TargetedTerritory, Trail}
-import org.joda.time.DateTime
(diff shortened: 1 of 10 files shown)


fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -110,14 +116,89 @@ object Collection {
...
+      maybeExpired.getOrElse(false)
+    }
+
+
+    val responses: Seq[Response[FaciaContent]] = faciaContent.map {
+      case curatedContent: CuratedContent =>
+        getMediaAtomData(curatedContent).map {
+          case mediaAtomData@Some(_) => curatedContent.copy(mediaAtomData = mediaAtomData)
+          case None => curatedContent
+        }      case content => Response.Right(content)
```
davidfurey, line 185:
```suggestion
        }      
      case content => Response.Right(content)
```

PR #357, opened 2025-06-02
Refactor mediaAtomData to mediaAtom on the facia content collection model

abeddow91:
## What does this change?
Refactor mediaAtomData to mediaAtom on the facia content curation collection model

This is so that we have access to the whole atom, rather than just the media atom data. The whole atom is required downsteam as the atom id is nested at the top of the model, not on the media atom. Without the ID, the media atom cannot be modeled by Frontend and passed to the rendering layer, DCR

This revision means we now validate in the consumer that the atom is of type media.   

This PR addresses some formatting so it is advised to review without whitespace changes to help reduce noise.

## How to test
Use the preview version of this branch and use it in Frontend. Validate that facia-press correctly adds a mediaAtom to any article that has either a main media or replacement video. 

## Accessibility

## Deployment

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -4,7 +4,7 @@ import com.gu.contentapi.client.ContentApiClient
 import com.gu.contentapi.client.model.v1.{Content, ItemResponse}
 import com.gu.contentatom.thrift.{Atom, AtomData}
 import com.gu.contentatom.thrift.atom.media.MediaAtom
-import com.gu.facia.api.contentapi.{ItemQueries, LatestSnapsRequest, LinkSnapsRequest}
+import com.gu.facia.api.contentapi.{ LatestSnapsRequest, LinkSnapsRequest}
 import com.gu.facia.client.models.{CollectionJson, SupportingItem, TargetedTerritory, Trail}
 import org.joda.time.{DateTime, DateTimeZone}
 import com.gu.facia.api.utils.BoostLevel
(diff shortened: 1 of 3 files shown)


fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -4,7 +4,7 @@ import com.gu.contentapi.client.ContentApiClient
 import com.gu.contentapi.client.model.v1.{Content, ItemResponse}
 import com.gu.contentatom.thrift.{Atom, AtomData}
 import com.gu.contentatom.thrift.atom.media.MediaAtom
-import com.gu.facia.api.contentapi.{ItemQueries, LatestSnapsRequest, LinkSnapsRequest}
```
abeddow91, line 7: Linting : unused import

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -155,19 +152,21 @@ object Collection extends StrictLogging {
...
+          if (!isExpired(mediaData.media)) {
+            true
+          } else {
+            logger.warn(s"Media atom is expired in ${id}")
+            false
+          }
+        case _ =>
+          logger.warn(s"No valid media atom found in ${id}")
+          false
+      }
```
abeddow91, line 167: We use the same case match for replacement or main media videos so I've extracted it here.

fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -118,35 +126,24 @@ object Collection extends StrictLogging {
...
-              case Some(_: AtomData.Media) =>
-                logger.warn(s"Media atom $atomId is expired")
-                None
-
-              case _ =>
-                logger.warn(s"No valid media atom found in CAPI response for ID $atomId")
-                None
-            }
+            response.media.flatMap(atom =>
+              if (isValidMediaAtom(atom, atomId)) Some(atom) else None
```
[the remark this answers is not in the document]
abeddow91: Thanks, I've updated to the `Option.when` syntax

PR #358, opened 2025-06-23
Main media selection bug

abeddow91:
## What does this change?

Adjusts the process for selecting a main media video atom by validating the atom ID against main media atom id.

This follows the pattern implemented in Frontend and resolves a bug where, if content has multiple atoms, the main media atom is not necessarily selected https://github.com/guardian/frontend/blob/7ac77e8bb7026b61f9712701c22db4e55e21919f/common/app/model/PressedStory.scala#L32

**before** 
Screenshot 2025-06-23 at 10 14 21

**after**
Screenshot 2025-06-23 at 10 14 05

## How to test
- Press a front on main with [this article](https://m.code.dev-theguardian.com/world/live/2025/jun/22/israel-iran-war-live-trump-says-us-has-attacked-nuclear-sites-in-iran-including-fordow). Check the DCR json and validate that the mediaAtom is *not* the main media. 
- Use the preview version of this branch in frontend and run locally. Validate that the mediaAtom *is now* the main media. 

Main media videos are being correctly attributed to the mainAtom field.

## Deployment

### DIFF
--- a/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
+++ b/fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
@@ -1,14 +1,14 @@
 package com.gu.facia.api.models
 
 import com.gu.contentapi.client.ContentApiClient
-import com.gu.contentapi.client.model.v1.{Content, ItemResponse}
+import com.gu.contentapi.client.model.v1.Content
 import com.gu.contentatom.thrift.{Atom, AtomData}
 import com.gu.contentatom.thrift.atom.media.MediaAtom
-import com.gu.facia.api.contentapi.{ LatestSnapsRequest, LinkSnapsRequest}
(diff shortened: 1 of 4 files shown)


fapi-client/src/main/scala/com/gu/facia/api/models/collection.scala
```diff
@@ -152,17 +154,29 @@ object Collection extends StrictLogging {
       Response.Async.Right(futureMaybeAtomData)
     }
 
-    def isValidMediaAtom(atom: Atom, id: String): Boolean = {
+    //
+    // We need to make sure that we only select the main media atom rather than another embedded atom. This follows the same pattern implemented in Frontend.
+    //
+    def extractMainMediaAtomIdFromHtml(html: String): Option[String] = {
```
davidfurey, line 160: I'm not sure why you need to parse the html to get the main media atom.
The CAPI response has this in response->content->blocks->main->elements->contentAtomTypeData->atomId

